PR #4, opened 2022-01-07
sync with gam-g4, HitsCollections

dsarrut:


tbaudier:
The tests are ok now with the version 0.3.0 of gam-g4 (with geant4 11) except for tests n° 19, 25 and 27.
Is it ok for merging?

dsarrut:
yes, its ok ! Thanks a lot. 
Test 19 is special, I should work a bit more on it.
Others tests 25 and 27 are still wip.

PR #9, opened 2022-02-25
Gam g4

tbaudier:
Integrate gam_g4 inside gam_gate

The history of gam_g4 is kept but the files are moved so it's not straight visible.
Some file are copie because the created conflict with the history: CMakeLists, .gitignore, .gitmodules, and files in .github/workflow
The submodules (fmt and pybind) are updated


tbaudier:
https://github.com/OpenGATE/gam-gate/issues/5

PR #17, opened 2022-03-07
Add version in one file

tbaudier:
To avoid to repeat version number, the version is written in the file VERSION and read by the setup.py file


dsarrut:
good idea!

PR #28, opened 2022-04-06
Add doc extension

tbaudier:
Add copy button for code area. You can copy the code more easily

Add a link to the source code for the API doc. I use this model https://github.com/aaugustin/websockets/blob/main/docs/conf.py#L102 but I cannot set the lines because gam_g4 is compiled here


dsarrut:
excellent !!!

PR #29, opened 2022-04-08
gagaga 

dsarrut:
GAN GAte source in GAm (GAGAGA)
- callback from c++ side to py side to trigger pytorch GAN particle generation (works with gpu!)
- still issued on osx -> seg fault after the end of the GIL (no seg fault on linux). workaround with a del
- not tested for multithread (yet)

See test 034


dsarrut:
ah, yes, sorry, the module gaga_phsp is missing

dsarrut:
arg, torch is required also ... unsure if it can be done here.

PR #37, opened 2022-06-03
Add default parameters to volumes store

BDoignies:
Hi, 

I pulled last version of gam gate and when running tests, I stumbled upon this error : 

```python
Running: test004_simple.py                          FAILED !   1.0 s     anonymized/gam_gate/gam_gate/gam_tests/log/test004_simple.py.log
Traceback (most recent call last):
  File "anonymized/gam_gate/gam_gate/gam_tests/src/test004_simple_MT.py", line 69, in <module>
    sim.initialize()
  File "anonymized/gam_gate/gam_gate/gam_gate/Simulation.py", line 198, in initialize
    self.g4_RunManager.Initialize()
  File "anonymized/gam_gate/gam_gate/gam_gate/geometry/VolumeManager.py", line 175, in Construct
    vol.construct(self)
  File "anonymized/gam_gate/gam_gate/gam_gate/geometry/VolumeBase.py", line 67, in construct
    self.construct_physical_volume()
  File "anonymized/gam_gate/gam_gate/gam_gate/geometry/VolumeBase.py", line 97, in construct_physical_volume
    mother_logical = st.GetVolume(self.user_info.mother, False)
TypeError: GetVolume(): incompatible function arguments. The following argument types are supported:
    1. (self: gam_g4.gam_g4.G4LogicalVolumeStore, arg0: gam_g4.gam_g4.G4String, arg1: bool, arg2: bool) -> gam_g4.gam_g4.G4LogicalVolume

Invoked with: <gam_g4.gam_g4.G4LogicalVolumeStore object at 0x7feb40696ab0>, 'world', False
```
I'm not sure of the "source" of the error, if it is because I have very new version of python (3.10) or install problems. Nonetheless, here is a quick fix of the error that enforces default values in python side.

### DIFF
--- a/gam_g4/gam_g4/g4_bindings/pyG4LogicalVolumeStore.cpp
+++ b/gam_g4/gam_g4/g4_bindings/pyG4LogicalVolumeStore.cpp
@@ -15,7 +15,7 @@ void init_G4LogicalVolumeStore(py::module &m) {
     py::class_<G4LogicalVolumeStore>(m, "G4LogicalVolumeStore")
 
         .def("GetInstance", &G4LogicalVolumeStore::GetInstance, py::return_value_policy::reference)
-        .def("GetVolume", &G4LogicalVolumeStore::GetVolume)
+        .def("GetVolume", &G4LogicalVolumeStore::GetVolume, py::arg("name"), py::arg("verbose") = false, py::arg("reverse") = false)
 
         // Additional functions, because the store is a vector
         .def("size", [](G4LogicalVolumeStore *r) { return r->size(); })
--- a/gam_g4/gam_g4/g4_bindings/pyG4PhysicalVolumeStore.cpp
+++ b/gam_g4/gam_g4/g4_bindings/pyG4PhysicalVolumeStore.cpp
@@ -17,5 +17,5 @@ void init_G4PhysicalVolumeStore(py::module &m) {
 
         .def("GetInstance", &G4PhysicalVolumeStore::GetInstance,
              py::return_value_policy::reference)
-        .def("GetVolume", &G4PhysicalVolumeStore::GetVolume);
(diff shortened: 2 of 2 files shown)


BDoignies:
The failed test from CI indicates that the error comes from newer version of geant4 that changed signature method.

tbaudier:
Hi @BDoignies 

What is the version tag fo Geant4 you use? Here, we use v11.0.0.
For python version 10, maybe we can encounter some problems because we do not use that version yet.

BDoignies:
For some reason geant4 switched to the master branch, which correspond to 11.0.2. I switched it back to v11.0.0 and the test still don't pass. 

https://github.com/Geant4/geant4/blob/geant4-11.0-release/source/geometry/management/src/G4LogicalVolumeStore.cc

As you can see, in the v11.0-release tag, there is those 3 parameters, although the CI tests indicates it does not.

dsarrut:
it is three params, but two lasts have default values in the include: https://github.com/Geant4/geant4/blob/geant4-11.0-release/source/geometry/management/include/G4LogicalVolumeStore.hh#L70
Your proposal is good if the default values are the same as the ones in G4.

BDoignies:
But python does not know about it...

As stated in pybind it can not extract default values : https://pybind11.readthedocs.io/en/stable/basics.html#default-args

That's what the PR does, making sure it can work. 

Also not that in the beta version of v11, only two parameters existed : 

https://github.com/Geant4/geant4/blob/6399a014b611a4aa09ecba225418222dac261ca1/source/geometry/management/src/G4LogicalVolumeStore.cc

BDoignies:
As for the default value, they are not quite right. Verbose parameter should be true by default (which is false in the PR). I'll update the PR

https://github.com/Geant4/geant4/blob/geant4-11.0-release/source/geometry/management/include/G4LogicalVolumeStore.hh
https://github.com/Geant4/geant4/blob/geant4-11.0-release/source/geometry/management/include/G4PhysicalVolumeStore.hh

PR #64, opened 2022-08-26
Change python version

tbaudier:
The version 3.6 of python is not supported anymore.
The version 3.10 was released a year ago


tbaudier:
https://endoflife.date/python

PR #70, opened 2022-09-12
PET gan voxelized source and speed test t045

dsarrut:
Warning, the t044 are writing files to the data folder, while they should only write to the output folder. To be corrected

Also the commit message "test044 voxelized GAN for PET" is wrong, should be "test045 voxelized GAN for PET"


dsarrut:
permission issue ? "sh: /Users/runner/hostedtoolcache/Python/3.7.14/x64/lib/python3.7/site-packages/opengate/tests/src/test045_speedup.py: Permission denied"

dsarrut:
Note  : it is not possible to directly call python function, Thomas, because it that case several G4 engines should be use in the same process, and it is not possible.

PR #85, opened 2022-11-14
Wheel for M1

tbaudier:
Create the wheel for Mac OS M1
Use Cirrus CI because M1 is not available on github Actions

Use cibuildwheel to create the wheel and easily manage the library on MacOS
Use that turnaround for delocate https://github.com/neuronsimulator/nrn/pull/1897/files

Push the artifacts on Pypi only for tagged version
Add cirrus CI badge in the readme

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,131 @@
+macos_instance:
+  image: ghcr.io/cirruslabs/macos-ventura-base:latest
+
+build_wheel_task:
+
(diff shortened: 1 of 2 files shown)


SimonRit:
FYI, there is a different approach followed by ITK for the ARM architecture, see InsightSoftwareConsortium/ITK#3471. This is not ready yet but could be a Github action alternative.

PR #88, opened 2022-11-18
Corrections

dsarrut:
This should have been several PR, but I fuckup my branches :( 
- correction of voxel source "in voxel" position (I add this bug in the previous commit)
- voxel source position (test021) updated
- replace postPosition with prePosition in HitsAdder
- towards "readout" actor (named "discretizer")
- a bit of doc
- test037 still in progress


tbaudier:
Do not take into account the error with the CI from cirrus ("pypi_push"). test021 branch contains the old version of the config file for cirrus with an error inside.

PR #91, opened 2022-11-23
beta_minus_spectra

jizhang02:
Hello,

Let me first introduce the background, suppose one wants to simulate the **Vivo radiation therapy**. The protocol is as follows:

- A voxelized source is added in the form of a dose map;
- "Lu177" is used as particles of source;

The test file is [here](https://gist.github.com/jizhang02/6778a73b5ba245aa404eb371584b049f).
According to [LNHB](http://www.lnhb.fr/nuclear-data/module-lara/) points out, "Lu177" is beta minus radionuclides.

However, the current code seems does not support it. 
Thus, the goal of this pull request is to **extend from beta plus to beta minus radionuclides**.
Thus, in this case, some minor changes are added in the file `GenericSource.py` and `helpers_source.py`. Please see the code below and consider if it is reasonable.

Also, there are two things that I'm not sure about.

1. Is `energy.type = "Lu177" ` the same as `energy.type = "spectrum"`?, I do not quite understand the format here.
2. In  `GenericSource.py`, line 122, it writes `# special case for beta plus energy spectra`, below it, I write `# special case for beta minus energy spectra`, but I mark it as `#TODO`, because I'm not sure how to implement, are they the same? Which is beyond my knowledge.

That's all, thank you!

### DIFF
--- a/opengate/source/GenericSource.py
+++ b/opengate/source/GenericSource.py
@@ -117,7 +117,7 @@ class GenericSource(gate.SourceBase):
                 gate.fatal(
                     f"Cannot find the energy type {self.user_info.energy.type} for the source {self.user_info.name}.\n"
                     f"Available types are {l}"
-            )
+                )
 
         # special case for beta plus energy spectra
         if self.user_info.particle == "e+":
@@ -133,8 +133,8 @@ class GenericSource(gate.SourceBase):
                 self.g4_source.SetProbabilityCDF(cdf)
 
         # special case for beta minus energy spectra
-        #TODO
-        
+        # TODO
+
         # initialize
         gate.SourceBase.initialize(self, run_timing_intervals)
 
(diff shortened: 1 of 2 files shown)


dsarrut:
Hello, 
for the moment, to simulate e- from ion decay, you can use an ion source (see for example [here](https://github.com/OpenGATE/opengate/blob/master/opengate/tests/src/test013_phys_lists_base.py#L78)). It means that G4 will simulate all the decay chain, including the e- (and e+ and gamma) emissions. It works well but can be long if we are only interested, for example, in e+. This is why the beta_plus_emitter system has been proposed. to only simulate e+ with energy spectrum corresponding to the ion decay. 

For e-, the alternative is to use a e- source with the energy spectrum of Lu77. Please, contact me per email, we can exchange about that ...

PR #92, opened 2022-11-25
small adjustments

dsarrut:
- doc img filename
- rename fct


dsarrut:
issue with numpy ??

tbaudier:
np.float was deprecated before and now it's removed:
https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

PR #101, opened 2023-01-03
Add python 3.11 support for github actions

tbaudier:


tbaudier:
OK for Linux
For MacOS, python is downloaded from conda and now, python 3.11 is not available yet : https://anaconda.org/anaconda/python

PR #102, opened 2023-01-03
Simulation engines

dsarrut:
Large refactor:
- SimulationEngine allows spawning a process to execute the simulation
- SimulationOutput for results (need to be pickable, no image for the moment)
- useful for notebooks and multiple simulations in the same script
- update to G4 11.1 (need updated data)


dsarrut:
ok, fail tests are only due to the G4 11.1 changes in the database, the data have been updated on gitlab. Can you update, please @tbaudier ?

PR #105, opened 2023-01-06
refactor PBS

dsarrut:
- rename to explicit PencilBeam instead of PB
- make GateSingleParticleSourcePencilBeam inherit from GateSingleParticleSource to separate implementations and avoid repetition


dsarrut:
I purposely cancel the workflow, because I know there are two issues:
- test044_pbs_rot_transl.py : one of the plane image is empty (500), is it expected ? 
- test044_pbs_weight.py : there is a large difference (factor of about 2) in the image, but it was not detected before because of a small bug in the boolean

PR #115, opened 2023-01-19
Try windows github actions

tbaudier:


tbaudier:
The wheel for windows is ok now.
Remaining to do:

- [ ] Build G4 with MT
- [ ] Opengate with MT
- [ ] Opengate with fork/spawn
- [ ] QT ?

PR #123, opened 2023-02-13
Treatment Plan Source

MartiDvrk:
**This pull requests includes:**

- Implementation of the Treatment Plan Source + tests
- Implementation of a basic beamline model object in helpers_beamline.py
- Implementation of functions and objects to handle RT dicom plans in helpers_rt_plan.py (mostly by David Boersma-IDEAL code)
- Exposure of the following G4 calsses: G4UserLimits, G4StepLimiter, G4ProcessManager.
- Interface of set_user_limits, implemented similarly to set_cut

**Treatment Plan Source implementation.** 

- The TPS is implemented by creating a Pencil Beam Source for each spot in the RT plan. 
- Position and direction of each PB source are defined similarly to the Gate9 implementation (https://github.com/OpenGATE/Gate/blob/develop/source/physics/src/GateSourceTPSPencilBeam.cc).
- The only difference is that the rotation of the source can also be initialized with an angle different to the gantry angle, by using the self.rotation argument. Rotation should be provided as scipy.Rotation object.

INITIALIZATION: 

- The TPS is initialized with a source name and a reference to the simulation object.
- Then, the user must provide: a beamline model (implemented in helpers_beamline.py) and the number of particles to simulate.
- The spots positions, energies and weights can be sets in two ways:

  1. 	with the function self.set_spots_from_rtplan(rt_plan_path) --> spots and gantry angle are automatically read from the dicom file
  2. 	providing an array of spots objects (implemented in helpers_rt_plan.py) and setting it with self.set_spots(spots)

  Note: backward compatibility with the Gate .txt Treatment plan file is possible. (see spots_info_from_txt() in helpers_rt_plan.py)
-  Optionally, the user can provide a rotation and a translation that will be applied to all the PB sources, resulting in a global rotation and translation of the TPS.
- The pbs sources are added to the simulation object when self.initialize_tpsource() is called.

TESTS:
Tests include verification of:

- absolute dose in water
- gantry roation angle
- optics in air of hbl and vbl beamlines
- validation of tps against pbs

 

Note: dicom plans examples and complete beamline models were not provided because of MA privacy policy. 

**User Limits implementation:**
- User limits can be now set by the user with the function 'sim.set_user_limits()'.
- The function takes as input:
	- volume to which the limits apply. Propagation to children volumes is implemented as well.
	- limit type. Available types are: "max_step_size","max_track_length","max_time","min_Ekine","min_range".
	- limit value (e.g. 0.8 mm)
	- list of particles to which the limit applies.
- Implementation in PhysicsEngine.py --> initialize_user_limits(self,tree) follows Gate9 (https://github.com/OpenGATE/Gate/blob/develop/source/physics/src/GatePhysicsList.cc) for the content
  and the 'set_cut' blueprint.
- No direct testing, but it is successfully used in the TPS tests (optics and optics_vbl)


dsarrut:
(Wait #130 from @nkrah before merging this PR)

nkrah:
> (Wait #130 from @nkrah before merging this PR)

Yes please. I am working on PR #130 and trying to sort out some segfault issues.

nkrah:
Hey folks, I am done with PR #130. I hope we'll be able to merge some time soon into the master. 
If you agree, I would suggest looking where this PR needs step limits and how to use them as implemented in PR #130. If you want, we can look at that together. We can also schedule a video call, if you want. Let me know

MartiDvrk:
Hi Nils,

Nice! I would be interested in knowing more about the PR, once it is merged.
You can also text me on my institutional email: ***@***.***

Best,
Martina

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PR #124, opened 2023-02-15
Arf doseactor stepsize

nkrah:
Three enhancements/additons in one PR (sorry, should have split that, but would have been a mess now). 

* Enhanced the ARF actor to score deposited energy
* Enhanced the DoseActor to score counts per voxel
* Add StepLimiter per region functionality. That required several changes in the RunManager logic and some newly bound G4 classes.


nkrah:
Please ignore this PR, but do NOT delete it. Refer to PR #130 instead which implements step limiter per region.

tbaudier:
Do not merge yet

PR #128, opened 2023-02-27
Update opengate_tests

BishopWolf:
tests are  not executed without this change


dsarrut:
Hi Alex ! 
I am unsure, why do you think tests are not executed? The current "if test" is to check that the data are downloaded.

tbaudier:
Hi Alex,

Exactly, the if test ckecked if the data were correctly downloaded before to execute the tests. If the data are not correct, it failed.
I close the PR, the CI failed too. If you encounter problems, send us an email or open an issue.

Thomas

Issue #129, opened 2023-03-02
Downloading / updating G4 data


nkrah:
I just banged my had head for a moment because my simulation would not start because opengate could not find the G4 data files. Sure enough, the files in core/opengate_core/geant4_data were outdated. I think, what had happened is: I pulled a new version of opengate into my local repo with new paths set in the g4DataPath dictionary in g4DataSetup.py. The simulation then looks for data in these paths, but does not find them. The check_G4_data_folder function, which is in __init__.py in opengate_core and thus called upon "import opengate_core" finds the geant4_data folder, but does not detect that the subfolders are outdated. Therefore, the G4 data is not updated. 

I propose to refine the check_G4_data_folder function to actually check the subfolders against the g4DataPath dictionary. 

If there are no objections, I will create a PR for this in the next days.

dsarrut:
yes indeed! The way the g4 data is managed is not optimal.  
However, a warning: the check is performed during the import, so if the check takes time, it may delay the starting phase of **all** scripts (currently, the import phase is already "slow", mainly due to itk I think). 
Please, go with the PR ;) !

nkrah:
Agree, but don't think the check takes time. Downloading data is slow, but that only occurs once in a while after an update. I'll implement and check.

nkrah:
Is there any reason not to get the data from the Geant4 installation which is required anyhow? Does it need to be downloaded separately? By looking up the Geant4 install, one would automatically align G4 and opengate in terms of G4 data. 
Opinions?

tbaudier:
Hi @nkrah 

You're right when you compile opengate. But when you install opengate on your compute with `pip install opengate` , Geant4 is not installed on your computer. You need to download the data.

nkrah:
I see. But don't you need Geant4 even if you install opengate via pip? Opengate uses Geant4 to run the simulation. Or does pip pull a pre-compiled Geant4?

tbaudier:
Inside the wheel of opengate_core, you can find the g4 librairies. But we cannot insert the data into the wheel because they are too big. This is why we prefer to download them at first import of opengate_core. It's not perfect because, the data are not removed correctly with `pip uninstall opengate`

nkrah:
OK, now I got the full picture. 

Is it possible to write a "hook" for the uninstall to remove the data? Possibly also when running an update via pip?

tbaudier:
I have a look and it seems not, eg:
[...]

PR #130, opened 2023-03-03
Implement step limiter (G4UserLimits)

nkrah:
This PR bundles a series of commits to implement step size limitation per volume (region) in opengate, plus some bug fixes and cosmetic improvements along the way. 

It required new binding for some Geant4 classes and a thin wrapper around the G4RunManager to expose essential protected members. Changes mainly appear in Simulation and SimulationEngine classes, in VolumeManager and VolumeEngine, as well as PhysicsManager and PhysicsEngine classes. 

Note 1: The initialization order in SimulationEngine.initialize was changed because initializing the step limiter requires an initialized geometry (G4RunManager.InitializeGeometry). Only then can G4RunManager.InitializePhysics be called. This has to do with the states handled by the G4StateManager. The G4RunManager.Initialize function would internally initialize geometry and physics in one go, which is incompatible with step limiter initialization. 
Please refer to G4RunManager.cpp and exampleB2b in the Geant4 source. 

Note 2: The step limiter is only handled correctly in single thread mode. Multithread mode would also require a wrapped RunManager (see note 1). As a temporary workaround, a check might need to be added to raise an error if a step limiter is requested in MT mode.

### DIFF
--- a/opengate/tests/src/test045_gan_phsp_pet_gan_helpers.py
+++ b/opengate/tests/src/test045_gan_phsp_pet_gan_helpers.py
@@ -12,11 +12,11 @@ from box import Box
 
 def create_pet_simulation(sim, param):
     # units
-    Bq = gate.g4_units("Bq")
-    BqmL = Bq / cm3
     mm = gate.g4_units("mm")
     m = gate.g4_units("m")
     cm3 = gate.g4_units("cm3")
+    Bq = gate.g4_units("Bq")
+    BqmL = Bq / cm3
 
     # main parameters
     ui = sim.user_info


nkrah:
I just saw that PR #123 also mentions UserLimits. It's probably worth cross-checking.

nkrah:
Do not merge yet. I will still extend and modify this PR.

tbaudier:
FYI:
- [ ] https://github.com/OpenGATE/opengate/issues/129 waits to resolve
- [ ] https://github.com/OpenGATE/opengate/pull/130 waits for Nils update
- [ ] https://github.com/OpenGATE/opengate/pull/123 To adapt when https://github.com/OpenGATE/opengate/pull/130 is merged

nkrah:
I just pushed a bunch of commits. The PR has thus significantly changed. I still need to correct a few things, so please wait until you attempt to merge into master.

tbaudier:
> I think tests are currently failing because the installation does not place the file geant4_particles.txt in opengate_core. I guess one has to instruct the setup script that this file is required. @tbaudier : how does that work?

Exactly, you need to add geant4_particles.txt in https://github.com/OpenGATE/opengate/blob/master/MANIFEST.in and the file will be added inside the wheel

nkrah:
> > I think tests are currently failing because the installation does not place the file geant4_particles.txt in opengate_core. I guess one has to instruct the setup script that this file is required. @tbaudier : how does that work?
> 
> Exactly, you need to add geant4_particles.txt in https://github.com/OpenGATE/opengate/blob/master/MANIFEST.in and the file will be added inside the wheel

Perfect. That's the info I needed. Let me finish some stuff and I'll push it.

nkrah:
Alright, we are almost there. I implemented production cuts using the new Region mechanism in this PR and added a test specifically for that (test053_production_cuts). It tried to make the Simulation.set_cut() method as backwards compatible as possible. 

Production cuts are handled according to the following logic: 

* If nothing is set, the list's defaults are used (1 mm)
* If a user sets a global cut value, this is applied for the specified particle. For the other particles, the cuts remain the default. 
* If a user specifies a cut for a certain particle in a certain region, this is set accordingly. The others retain the global cut potentially set by the user, or the default if no global cut is set. 

For the user, it makes NO difference in which order these values are set. This is handled internally by the PhysicsEngine, namely: first initialize physics lists, then global cuts, then regional cuts (so the regional cuts can pick up global cuts set by the user). 

Important note: 
The G4RunManagerKernel::InitializePhysics() function calls physicsList->SetCuts() at the end, thereby applying the default cut value to all particles (proton, e+, e-, gamma). Therefore, if the PhysicsEngine initializes global and regional cuts before the RunManager initializes physics, the global cuts will be overridden. One can either manually call SetCuts() once before (because a flag is set internally), or initialize the cuts after the RunManager has initialized physics. I have so far implemented the latter, but the former is maybe better. I need to check.

nkrah:
The initialization order was incorrect. The physics engine must initialize user limits before the RunManager calls InitializePhysics because user limits are added to the physics list and that must happen at pre_init state. 

Also: The simulation is now forced to fall back to single thread if MT is requested, rather than triggering a fatal error. This is a temporary work around to avoid CI from failing due to this known non-implemented feature.

nkrah:
Update: 
* UserLimits and regional cuts now work in MT mode as well
* Processes needed for UserLimits (StepLimiter and UserSpecialCut) are added to the physics list via a PhysicsConstructor implemented in gate (python) in analogy to the StepLimiterPhysics constructor in G4. It gives more control to the user to which particles the cuts should be applied, but reproduces the behavior of StepLimiterPhysics by default. So, if the user sets no user_limits_particles[XXX] to True, a StepLimiter will be applied to all charged particles (The StepLimiterPhysics behavior). If the user sets, e.g. user_limits_particles['gamma']=True and user_limits_particles['all_charged']=False, ONLY the gamma will get the StepLimiter process, but not the charged particles. 
* As per previous commits in this PR, production cuts by region are handled via the Region object
* I began implementing a context manager logic for the SimulationEngine by added __enter__ and __exit__ methods. This is used via Simulation.run(). Output is stored in Simulation.output. the method run() uses a 'with' statement to create a context within which the SimulationEngine is created. Before existing, the SimulationEngine.__exit__() method is called which calls the close() methods newly implemented for all Engines. They take care of releasing references to G4 objects which will be deleted by the RunManager upon destruction. In this way, segmentation falls due to empty addresses at the end are prevented. 
* The above context mechanism is not (yet) compatible with the way simulations are moved into new processes, at the run() method currently works around this. A proper solution will require slight restructuring of the process handling, but that should go in a separate PR. 
* I homogenized the references in the Engine classes. Specifically, they are all set up now in a way that their __init__() methods take only the SimulationEngine as argument. All references to sister engines or managers are deduced from there, e.g. self.simulation_engine.volume_engine or self.simulation_engine.simulation.physics_manager. That makes the code slightly lengthy in certain places, but I think it is really important to have a consistent structure and hierarchy of references between engine and managers, especially for more advanced process handling in the future. 
* I also extended the two tests to alternatively test serial mode and MT mode, as well as execution in separate process of not. 
* I added a new option to GateObject, which only Region uses as base class for now, namely 'expose_items'. It can be seen to true if a user_info item is a dictionary (or similar) itself, e.g. user_limits. If 'expose_items'=True for user_limits, that also the individual items are exposed as properties. 

```
r = Region(name='R1')
# You can do this
r.user_limits.max_step_size = 10
# or this
r.max_step_size = 10
```
... both stock access the same user_info internally, but it's simpler for the user. 

From my side, this PR should be ready now. Let's look over it, see what tests fail, and then think about merging.

nkrah:
I should add: 
* I did not need the wrapped version of the G4RunManager anymore, but kept that binding in opengate_core nonetheless. Maybe it will be useful in the future. 
* I also abandoned using the RunManagerFactory now and have created a method in SimulationEngine instead to select the proper type of RunManager according to user input and to what is available (or fail). The RunManagerFactory does nothing else, essentially, plus it checks some environment variables which G4 users can set, but which might be confusing Gate users.

nkrah:
@dsarrut @tbaudier I went through all tests and updated them to work with the new way of setting production cuts and running simulations. I fixed some minor bugs along the way. A few tests continue to fail. 

- In some cases because of slight deviations of the statistics. 
- In other cases because images do not match. 
- I did not run tests locally (yet) which required PyTorch. 
- Some of visu tests have stats that are off by orders of magnitude (at least on my local machine). Maybe because the number of primaries is strongly reduced?

Is there a way to get a simple list of those tests that failed without going through the log in "Actions" here on GitHub?

I will double-check the still failing tests. Maybe we can check the persistent ones together? 

Otherwise, the PR is good to go, and we can begin considering a merge into master, I think.

nkrah:
Update: 
tests in CI timed out. I downloaded the wheel, installed it locally in a proper environment. Tests got stuck at test023_filters_iec_phantom.py at the end of the simulation.  
Strange thing: the simulation runs fine within the src directory of my local code, i.e. which uses directly my local python code, not the package installed from the wheel. 

I will do a full local run of tests over night and see what comes out.

nkrah:
Another update: 
CI just now failed because of tissue after updated colored package (see issue #176 ).

nkrah:
Provided a fix for broken syntax after updating colored package to v1.5

nkrah:
Implemented fixes in parallel world handling. 

They were due to that part being implemented in parallel to structural changes in this PR.

nkrah:
Locally, i.e. on MacOS, python 3.9, all tests pass.

PR #132, opened 2023-03-07
Improve G4 data setup

nkrah:
Changed and extended the Geant4 data setup. Now the script not only checks whether the geant4_data folder exists, but also checks whether all required subfolders exist. If anything is missing, a full fresh download of all data is triggered. 

The script introspects the tar files to get the directory into which the tar files are extracted and in this way identifies subfolders which are probably outdated and not needed anymore. The user is informed about this, but the files are not removed. 

If a subfolder into which a tar file needs to be extracted already exists (e.g. from a previous failed attempt to install G4 data), this script first deletes the directory and lets the tarfile library recreate it. This has to do with the way user IDs and permissions are handled by tarfile. 

NB: The PR also corrects some improperly defined dictionaries (env variables) and adds an environment variable for a dataset that was previously missing.


nkrah:
Some background info about tar extraction: The tar file contains the user ID of the user who packaged the tar, e.g. somebody at CERN. This user ID does most likely not correspond to the local user ID on the machine where opengate is installed. extractall function of the tarfile library uses the user ID given by the tar file to extract into a directory and changes the user IDs to those of the local user running the python script only after the extraction process. This means, if the directory into which the files are extracted already exists, e.g. from a previous run of the script, and it does not give write access to everybody, most likely the extractall function will fail because of lack of permission. If instead the directory does not exist yet, extractall creates it (with the user ID from the tar file) and the extraction works fine. Therefore, the download_G4_data() function in this PR checks whether the desired destination directory already exists and deletes it in that case before running the extractall() function. 

Here is the excerpt from tarfile's doc: 
"Extract all members from the archive to the current working directory or directory path. If optional members is given, it must be a subset of the list returned by [getmembers()](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.getmembers). Directory information like owner, modification time and permissions are set after all members have been extracted. This is done to work around two problems: A directory’s modification time is reset each time a file is created in it. And, if a directory’s permissions do not allow writing, extracting files to it will fail."
[https://docs.python.org/3/library/tarfile.html]

tbaudier:
Thank you very much Nils, I merged

nkrah:
Thanks. I was actually still slightly improving the PR. How should we do?

dsarrut:
Please also measure the computation time needed for the check. (a simple "import opengate" before and after the PR merge). itk import is already "very" slow. This is not an issue when running a large simulation, but it may become cumbersome when developing and running scripts. thanks !

PR #134, opened 2023-03-08
G4 data setup - improvements

nkrah:
Here are some improvements that I had not implemented yet before PR #132 has been merged. 

I implemented two functions to structure the code better, namely to check whether all required folders in geant4_data are present (used to trigger download) and to check whether there are folders or files in geant4_data that are no longer required. This functionality was already in PR #132, but the code structure is better now. 

I also replaced ugly loops by much cleaner operations on sets.


dsarrut:
Please also measure the computation time needed for the check. (a simple "import opengate" before and after the PR merge). itk import is already "very" slow. This is not an issue when running a large simulation, but it may become cumbersome when developing and running scripts. thanks !

nkrah:
Computation time is not an issue. 
"import opengate" only triggers the function check_g4_data() which only compares the existing directories in geant4_data against those in the environment path dictionary. That's a millisecond operation. I've checked that. 

Downloading and extracting the data takes time, but that has nothing to do with the PR. 

I don't see any issue here.

PR #138, opened 2023-03-20
Physic list bug

dsarrut:
From: @MartiDvrk
There is a strange behaviour with some physic lists that lead to zero output. 

"We have a suspicion that is related to “G4TENDL1.4”, which is new since the last or previous G4 release. We were a bit slowed down by having compiling issues since we pulled last days from the main branch. seems to be solved now.
I looked into the G4 examples and the G4 source code of the phys lists. There was nothing obvious, why it should work for BIC, but not for BIC_HP, which inherits from BIC and overloads few functions. On the other hand “Shielding”, does not inherit from another phys list, but also doesn’t work (has the HP inside)."

This is an example that shows the bug. It may be related to the "G4RadioactiveDecayPhysics". When it is enabled it fails. 
- QGSP_BIC do not enable G4RadioactiveDecayPhysics: it is correct
- QGSP_BIC_HP enable it: it fails. 

If we force QGSP_BIC_HP to remove the process (using RemovePhysics), it works.

### DIFF
--- a/core/opengate_core/g4_bindings/pyG4VModularPhysicsList.cpp
+++ b/core/opengate_core/g4_bindings/pyG4VModularPhysicsList.cpp
@@ -72,5 +72,9 @@ void init_G4VModularPhysicsList(py::module &m) {
            py::overload_cast<G4int>(&G4VModularPhysicsList::GetPhysics,
                                     py::const_),
            py::return_value_policy::reference)
-      .def("RegisterPhysics", &G4VModularPhysicsList::RegisterPhysics);
+      .def("RegisterPhysics", &G4VModularPhysicsList::RegisterPhysics)
+      .def("RemovePhysics",
+           [](G4VModularPhysicsList *s, G4VPhysicsConstructor *p) {
+             s->RemovePhysics(p);
+           });
 }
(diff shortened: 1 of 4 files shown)


dsarrut:
This is a workaround for testing. It is probably *not* the right way to do. 
Here, the user can force to REMOVE the process G4RadioactiveDecay even if it has been set by a physic list such as QGSP_BIC_HP. See:  [here](https://github.com/OpenGATE/opengate/pull/138/commits/214998a91ea655793dc7b1ee04c4031964dab4a1#diff-8c00d12ae11230a0d31f420b5042655019eaafbb9781e2338f0a518dab476909R70)

nkrah:
Can you check if the underlying issue described above is still valid with the current master version of Gate and Geant4 v11.1.1?
Thanks

nkrah:
Please follow issue #197 which investigates the problem which prompted this PR. 

DO NOT MERGE THIS PR - it is just a workaround.

nkrah:
PR #218 fixes the problem underlying the workaround proposed by this PR.

nkrah:
Leave this open for now for testing.

PR #148, opened 2023-04-12
Tps no step limiter

MartiDvrk:
Same as Treatment Plan Source (PR #123) but without StepLimiter. Test which required step limiter are for now passing with very high thresholds.


tbaudier:
Done in https://github.com/OpenGATE/opengate/pull/189 with a rebase.
I close

PR #154, opened 2023-04-25
fix initialize_physics_list 

m-dupont:
Changes comes from geant4 example where using `G4EmStandardPhysics_option4 `  is always preceded by instantiating FTFP_BERT and use method `ReplacePhysics` with  `G4EmStandardPhysics_option4 ` as argument. 

This will also allow us to add `G4OpticalPhysics` (wich can not be used directly on  `G4EmStandardPhysics_option4 `).

### DIFF
--- a/core/opengate_core/g4_bindings/pyG4VModularPhysicsList.cpp
+++ b/core/opengate_core/g4_bindings/pyG4VModularPhysicsList.cpp
@@ -72,5 +72,6 @@ void init_G4VModularPhysicsList(py::module &m) {
            py::overload_cast<G4int>(&G4VModularPhysicsList::GetPhysics,
                                     py::const_),
            py::return_value_policy::reference)
-      .def("RegisterPhysics", &G4VModularPhysicsList::RegisterPhysics);
+      .def("RegisterPhysics", &G4VModularPhysicsList::RegisterPhysics)
+      .def("ReplacePhysics", &G4VModularPhysicsList::ReplacePhysics);
 }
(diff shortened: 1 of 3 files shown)


m-dupont:
strange, the `test037_pet_hits_singles_1.py` successfully passed on my computer

tbaudier:
Hello Mathieu,

I had the error on my computer.
I do not know if it's related but for the `test037_pet_hits_singles_1.py`, I observed that:
Here: https://github.com/OpenGATE/opengate/pull/154/files#diff-8c00d12ae11230a0d31f420b5042655019eaafbb9781e2338f0a518dab476909R55 the type of `self.g4_physic_list` is `opengate_core.opengate_core.G4VModularPhysicsList` instead of `opengate_core.opengate_core.G4EmStandardPhysics_option4` before the commit.

I have no idea if it's related or not

nkrah:
Thanks Mathieu, 
that's a good point with the G4OpticalPhysics. In fact, the G4EMStandardPhysics is simply a G4PhysicsConstructor, not a G4ModularPhysicsList. The latter is essentially a container which contains several G4PhysicsConstructor objects. Gate already the uses the kind of "trick" you mention from the G4 examples, namely it creates a physics list object and assigns the single physics constructor, e.g.  G4EMStandardPhysics to it. It is therefore straightforward to additionally assign other constructors as well such as optical physics. 

We have recently merged a PR into the code which introduces a new PhysicsListManager. I will pick up your suggestion from this PR and implement that in that context, if that is OK for you?

nkrah:
The changes proposed in this PR have been considered in PR #187 which has recently been merged. Therefore, I am closing this PR. 
Thanks to Mathieu for the suggestions!

PR #181, opened 2023-06-21
Be sure to have a version of colored>1.5

tbaudier:
Avoid to have a try and except system to manage package versions. We prefer to set a minimal package version


tbaudier:
The newest version of colored is not available for python 3.8

dsarrut:
OK, we are waiting for end of life python 3.8 before merging (end 2024)

tbaudier:
Done here https://github.com/OpenGATE/opengate/pull/457

PR #188, opened 2023-06-29
Em parameters

nkrah:
This PR implements EM parameter switches such as fluo, auger, auger_cascade etc. So far (test013_phys_list_helpers), these parameters could only be set by directly interacting with G4 objects. 

I took this occasion to turn the PhysicsManager into a child of GateObjectSingleton. That guarantees the only one physics manager exists and it implements a new way of handling user_infos, as already the case in the Region class. 
This switch makes the PhysicsUserInfo class obsolete. 

EM parameters are implemented as boolean switches, but they are set to None by default. None means that Gate does not touch the EM parameters and leaves it to G4 to use the model's defaults. Deexcitation in the world region is activated via user infos in the physics manager.  Deexcitation in specific regions is activate via the respective region object, i.e. via user infos in the Region class. 

The method PhysicsEngine.initialize_em_parameters communicates the parameters to G4. Region deex activation is communicated via a similar initialize method in the Region class. 

Caveat: The current (master) tests test013_phys_list_2 and test013_phys_list_5 produce inconsistent data, and ref data for these tests is most likely flawed. In fact, G4 verbose output shows that the deexcitation is never correctly initialized in these tests, i.e. G4UAtomicDeexcitation::InitialiseForNewRun() is never executed correctly. 
Further, G4 output shows 
Ignore cuts flag:   0
although the test scripts set it True.  
I think this incorrect behavior in the current tests comes from the fact that the EM parameters are set manually via the G4EmParameters.Instance() *before* the physics list is created. if the EM parameters are set after creating the physics list,         self.g4_RunManager.SetUserInitialization(self.physics_engine.g4_physics_list) in SimulationEngine, the G4 output reports correct flags. 

The current PR by construction sets the EM parameters after the physics lists are created and G4 output shows consistent flags. However, with Ignore cuts flag:   1, as requested by the tests, the statistics are naturally much higher than in the (inconsistent) ref data. For this reason, I marked tests test013_phys_list_2 and test013_phys_list_5 as WIP.


nkrah:
I now implemented two special user hooks which are executed after G4 has run the simulation. They check the EM parameters (flux, auger, etc.) as well as the regions in which deexcitation has been activated. The mechanism is slightly complicated because the flags on the G4 side are declared private. The G4VAtomDeexcitation class provides two check functions, but they take as inout the cuts couple index. So this needs to be determined first for each region. 

The  hooks are in opengate.user_hooks and use a function implemented in opengate_core/pyGateCheckDeex.cpp

nkrah:
Test060_PhsSource used obsolete sim.start() instead of sim.run() -> was updated. 

In test061, enable_decay was not set correctly yet via the new PhysicsManager -> test failed. now updated and OK.

nkrah:
For info: 
test058_calc_uncertainty_MT.py was failing because of a TypeError raised by ITK. Specifically: 
TypeError: in method 'itkImageIOFactory_CreateImageIO', argument 1 of type 'char const *'

After updating ITK (via pip) from 5.2.1.post1 to 5.3.0, the error is gone.

nkrah:
> For info: test058_calc_uncertainty_MT.py was failing because of a TypeError raised by ITK. Specifically: TypeError: in method 'itkImageIOFactory_CreateImageIO', argument 1 of type 'char const *'
> 
> After updating ITK (via pip) from 5.2.1.post1 to 5.3.0, the error is gone.

similar for test059 -> resolved after ITK update

nkrah:
The last commit made GateObjects compatible with sub processing using "spawn" (necessary for Windows). In fact, a recent PR had switched Gate's subprocessing from "fork" to "spawn" which broke the user_info handling in GateObjects. I will write up a little section in the dev docs,  but in short: 

The python sub processing pickles objects and unpickles them in the subprocess. When using "fork", this happens within the same python interpreter scope, while "spawn" triggers a new python interpreter. The GateObject class (from which Region, PhysicsManager, PhysicsListManager inherit) relies on a mechanism to digest all user_infos defined by all classes along the line of inheritance. This mechanism is triggered when the first object instance of the specific class is created. The unpickling, by default, does not trigger this mechanism and therefore the classes were lacking the properties associated with user_infos, which in turn made the code fail in many places. In "fork" subprocessing, the problem was masked because the classes had been processed once already (and stored in the meta class) within the scope of the same python interpreter. 

The commit solves this by implementing a __reduce__() method for GateObject and GateObjectSingleton which is called  during pickling and which specifies how the picking/unpickling should take place. In particular, we can provide a function (a callable) which triggers the user_info digestion mechanism of a class before any object is created.

PR #195, opened 2023-07-07
TP source with step limiter 

MartiDvrk:
Added step limiter to the optics tests and lowered the pass threshold.


tbaudier:
Is it possible to reduce the time of the test test059_TPSource_abs_dose.py?
Now, it lasts ~2min30. Could we reduce the time to less than 1min?

tbaudier:
Thank you very much, I merge

PR #200, opened 2023-07-13
Update IEC phantom

dsarrut:
The IEC was described with some unions of solids. 
There was a slight position error, that may create strange behaviour with G4 error such as below. 
I am not fully sure the union is perfectly working now, but at least I did not encounter this error. 

```

	G4WT0 > 
-------- WWWW ------- G4Exception-START -------- WWWW -------
*** G4Exception : GeomNav1002
      issued by : G4Navigator::ComputeStep()
Stuck Track: potential geometry or navigation problem.
  Track stuck, not moving for 10 steps.
  Current  phys volume: 'iec'
   - at position : (69.99999999999999,-2.016610842758837e-06,-28.55691939210681)
     in direction: (0.9324614260989251,-0.2578523273110441,-0.2530372821104372)
    (local position: (69.99999999999999,-2.016610842758837e-06,-28.55691939210681))
    (local direction: (0.9324614260989251,-0.2578523273110441,-0.2530372821104372)).
  Previous phys volume: 'world'

  Likely geometry overlap - else navigation problem !
   *** Trying to get *unstuck* using a push - expanding step to 1e-07 (mm) ...       Potential overlap in geometry !
```


dsarrut:
error

PR #202, opened 2023-07-13
Upgraded phs source

GitFuchs:
The phase space actor was extended to also store the PDGCode in the root file. 
This makes evaluations potentially much more efficient.

The phase space source was extended to be able to translate and rotate the source.
Furthermore, it can now use the particle name or PDGCode from the phase space. This way also secondary particles can be simulated.

Tests were created.
test060 for the phase space source
testtest019_linac_phsp_PDGCode for the phase space actor


dsarrut:
great thanks ! Test037 failed: add one more attribute (40 instead of 39) ;)

GitFuchs:
Fixed it :-)

tbaudier:
Perfect, thank you I merge

PR #205, opened 2023-07-21
Tps phs source

GitFuchs:
I have created a new source, the TreatmentPlanPhsSource.
It is a near copy of the TreatmentPlanSource.
Instead of a beam modell, it uses phase spaces for the single energies. E.g. a list of phase spaces.
Test061_TPPhsSource was created.
Instead of a beam modell one needs a text file linking the energy to the corresponding phase space file and of course the phase space files themselves.

The source has the same limits as the Phase Space Source, that is it at the moment is not multithreading capable.
Data for the test to be put in data/test061_ref was sent to TBaudier for inclusion.


tbaudier:
I added the data and did the modification in the test to deal with the new path.
But there is still an error:
Inside the file data/output_ref/test061_ref/PhsList.txt there is the line:
150.0	../data/output_ref/test061_ref/test_proton.root
But the path is relative and lead to error because it depends where the tests is running.
Is it necessary to write the path in an external file? Could we imagine to set in the python test directly?

GitFuchs:
The phase space source was extended a bit to allow simulations until the next primary particle. Leftover code in opengate/source including the now empty folder was removed.
The failing tests060 were fixed.

I moved opengate/sources/treatmentPlanPhsSource.py to contrib/tps 
and fixed the associated test061
The TreatmentPlanPhsSource does similar things than the TreatmentPlanSource from opengate.contrib.tps.ionbeamtherapy, but instead of using a parametric beam model, it reads in phase space files. This should allow more complex particle spectra as well.

Now the phase space source, as well as the TreatmentPlanPhsSource should be up to the new Gate structure.

GitFuchs:
test060_phsp_source_untilNextPrimary.py requires pandas to create a root file. Apparently on the test wheel this is not installed.
Is this a critical issue?

GitFuchs:
The dependence on pandas of test060_phsp_source_untilNextPrimary.py was removed

PR #217, opened 2023-08-10
Creation of a kill actor which killed each particle crossing it

majacquet:
Creation of a kill actor. Don't merge it now, I pushed to obtain the wheels


dsarrut:
#263

PR #225, opened 2023-08-25
Spawn process (instead of fork) + ARF adaptation

dsarrut:
- Replace "fork" by "spawn" when a new process is created : all script **MUST** start with 
`if __name__ == "__main__":`
- ARF prediction now uses cpu by default (better for scalability, GPU is not faster here)
- MotionActor has been changed : open geometry must only be done in Master Thread
- all tests updated


dsarrut:
Tests with ARF failed because garf wheel need to be updated

PR #229, opened 2023-09-05
Phsp source : avoid pybind copy

dsarrut:
In phsp source, the particles are generated from python. Now, we avoid copying the data from python to cpp, it should be faster. 
We also change the default uproot awkward to numpy because it was not working on linux (but was fine on MacOS, dont know why). 
Still the MT version is not really faster. Maybe with simulation where the tracking is larger ?


dsarrut:
@GitFuchs : I had to change a bit test060 and remove the possibility of using "ParticleName" from the phsp. This is because, now, the data read from the root file by python are not copied to the cpp side but shared. It is faster, but I am unsure how to do that for string type. Can you please check that all tests are still ok for you please ? If ParticleName is important to you, we should find a way to add this option.

PR #233, opened 2023-09-11
Init gateobjects

nkrah:
Small PR which makes sure that classes inheriting from GateObject and GateObjectSingleton are processed when the opengate module is imported. This makes sure, e.g. that the docstring is already complete, and auto completion in interactive python consoles such as ipython should also pick up user_infos. 

Alongside, the implementation style in GateObjects.py is also improved, defining methods like __init__, __new__ etc explicitly now, rather than programmatically, as before. 

IMPORTANT: This PR is based on PR #188, so it should only be merged after that PR!


nkrah:
Already included in PR #186

PR #238, opened 2023-09-15
Dose let actor MT

andiresch:
- added dose to water to DoseActor, including test case
- made Dose to water and LETActor, MT compatible, adapted test case to test with MT
- converted 4D image in Dose actor to ThreadLocal Vector
- changed DoseActor to only have one output variable (edep, dose, dose to water), plus uncertainty
- added option to score dose map for each thread in thread local vector; no time calculation improvements though
- added option to calculate the standard error from the above copy of images ; two test cases for it; 
- added particle filter to LETd test


andiresch:
this pull request is now obsolete and is fully integrated into PR #293 and shall be deleted.

PR #241, opened 2023-09-19
Consolidate

dsarrut:
- add doc for phsp source 
	- correct vrml default filename
	- try to write "empty" root file during phsp actor, but I cannot manage. Add a warning instead
	- add enable_decay to Simulation object


nkrah:
Not sure about the fourth point. 
You can always do: 
sim.physics_manager.enable_decay  = True
just as you can do 
sim.physics_manager.phys_list_name = "xxx"
or all other user_infos in physics_manager

If we think that we need to expose "enable_decay" at the level of the Simulation object, we should make it a property, not a function.

dsarrut:
Yes agree : I think a property is better. 
I feel that if set_production_cut is accessible from Simulation, phys list and decay should be too, no ?

nkrah:
I think we will need to make a design choice, namely whether we want options and methods to be accessible via the managers or directly via the Simulation object. 

The first choice underlines the different aspects of a simulation, i.e. geometry, physics, sources, etc. 
The second choice lets the user type a bit less. 

Once this choice is made, I agree that it should be implemented consistently. 

Personally, I tend to prefer the first choice, i.e. pass through the managers, because it is more in line with the underlying Geant4 logic. Users will likely do copy/pasting or have an editor with auto completion (advantage of python over old macro files) so the shortcut is less relevant. Let's discuss this ...

PR #252, opened 2023-10-13
misc consolidation

dsarrut:
- rename tests to lowercase
- add voxelized source for IEC 
- update garf and gaga to new gpu model


tbaudier:
The garf and gaga update was done here:
https://github.com/OpenGATE/opengate/pull/249

But there is still a problem with python3.8 and MacOS

PR #262, opened 2023-10-19
Reduce import time of opengate

tbaudier:
It seems that import itk by module, increase a lot the time of import for opengate

I did this:
```
python -X importtime testTime.py 2> import.log
tune import.log
```
with:
```tesTime.py
import opengate
```
With my MacOS and python 3.9

Before the modification:
it took 11s
Screenshot 2023-10-19 at 16 34 22
Screenshot 2023-10-19 at 16 34 39

After the modification:
it tooks 5.2s
Screenshot 2023-10-19 at 16 48 46
Screenshot 2023-10-19 at 16 48 58

ITK is still long to import due to torch. But it's import just once now

### DIFF
--- a/opengate/actors/arfactors.py
+++ b/opengate/actors/arfactors.py
@@ -1,9 +1,7 @@
 import sys
 from box import Box
 import numpy as np
-from itk import image_from_array, Image, CastImageFilter, imwrite
-from itk import F as itkF
-from itk import D as itkD
+import itk
 import threading
 import opengate_core as g4
 from ..utility import g4_units, check_filename_type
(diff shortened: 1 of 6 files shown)


SimonRit:
When you do `import itk`, importing actually occurs when you start using it. This is the so-called lazy loading (which can be disabled with
```Python
import itkConfig
itkConfig.LazyLoading = False
import itk
```
). So I agree, with this change, there is no benefit in pre-loading things that you might not use eventually. ITK long loading time is a known isssue, a possible workaround would be to have a Gate specific itk package with only the relevant image types.

nkrah:
Interesting Simon, I did not know that. But good to know indeed!

PR #265, opened 2023-10-24
Filter any attribute with min/max

dsarrut:
A generic filter (min/max) for any attribute with double value.
Example : 

```
ka = sim.add_actor("KillActor", "kill_actor2")
ka.mother = plane2a.name
att_filter = sim.add_filter("ThresholdAttributeFilter", "ene_filter")
att_filter.attribute = "KineticEnergy"
att_filter.value_min = 300 * keV
att_filter.value_max = 1200 * keV
att_filter.policy = "keep"
ka.filters.append(att_filter)
```
In this case, all particles within 300-1200 keV are killed.

### DIFF
--- a/core/opengate_core/opengate_core.cpp
+++ b/core/opengate_core/opengate_core.cpp
@@ -237,7 +237,7 @@ void init_GateVFilter(py::module &);
 
 void init_GateParticleFilter(py::module &);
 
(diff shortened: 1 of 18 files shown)


dsarrut:
something strange with test51 on windows : maybe related to separate process ?

PR #278, opened 2023-11-02
UC Davis optical physics for merging

nkrah:
Branch created from PR #254 for easier merging with master.


nkrah:
The test which fails in this PR is unrelated to the changes in this PR. I am merging the PR because it is needed by the Gate course during IEEE-MIC. We will take care of the failing test afterwards.

PR #281, opened 2023-11-07
Json export step1

nkrah:
First steps of implementing a mechanism to dump/load a simulation as a JSON file.

### DIFF
--- a/core/opengate_core/opengate_lib/GateHelpers.cpp
+++ b/core/opengate_core/opengate_lib/GateHelpers.cpp
@@ -7,6 +7,7 @@
 
 #include "GateHelpers.h"
 #include <G4Threading.hh>
+#include <stdexcept>
 
 const int LogLevel_RUN = 20;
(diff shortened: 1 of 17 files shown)


nkrah:
JSON export works for those parts of the simulation which already make use of the GateObject base class. Those are: VolumeManager and volumes, PhysicsManager, PhysicsListManager, Region, and now also Simulation. 

The Simulation class has been refactored into a GateObject. User infos are now defined via the `user_info_defaults` class attribute. The SimulationUserInfo class has thus become obsolete and has been removed. 

Using the json archiving option is simple, see test065_sim_as_dict.py. 
The user activates it by setting the `store_json_archive` option in the simulation object to True. 
When running the simulation, a json file containing all relevant parameters of the simulation will automatically be created and stored in the output directory which can be specified via the `output_dir` option of the simulation object. 
The user can also manually call `sim.to_json_file()` to save the simulation as json file.

PR #287, opened 2023-11-15
Source activity

majacquet:
Update of the PhaseSpaceSource in order to use it with source.activity, therefore enabling the simulation of several runs with only one simulation.


tbaudier:
Done in https://github.com/OpenGATE/opengate/pull/300

PR #293, opened 2023-11-17
TP source and Dose actor new features

MartiDvrk:
**Treatment Plan Source:**
- the Treatment Plan Source is suppose to handle a single beam. The user can now specify the beam number of the beam to simulate when reading the beamset info from the treatment plan (either in dicom or txt form).
- flat generation flag. By default, the TP source will generate for each spot a fraction of the requested number of particles for the beam, proportional to the msw of the spot. If the flat_generation flag is enabled, each spot will get the same number of particles, while the weighting will be applied on the energy deposited by the particles for that spot. This feature was already existing in Gate9. A dedicated test was created, I can send you the .txt plan file for it.

**Dose Actor:**
- Integration of PR #238 from Andreas.
- added possibility to calculate dose (or dose to water) at the end of the simulation, by dividing edep (or edep to water) by the mass. In case the dose actor is attached to a  ImageVolume, a mass image is calculated from the CT (but the resolution of the dose actor has to be the same of the CT). For now, the flag to enable/disable this option is colled dose_calc_on_the_fly. To be set to False for dose calculation at the end of the simulation instead of in the stepping action. Probably there could be a better name.
- tests for the dose_calc_on_the_fly flag are included in the PR.
- added possibility to set a goal uncertainty, after which the stop of the simulation is triggered. This option works by calculating the uncertainty from the images of all threads at the end of each run, in the EndOfRunActionMasterThread. When the uncertainty is lower than the goal one, the IsEndOfSimulationForWorker flag in each SourceManager becomes true. As the uncertainty is calculated using the thread images as samples, this option only makes sense in multi-threading mode. An error is raised otherwise. Moreover, it is up to the user to set an appropriate number of runs. An example of use is included in test066.


nkrah:
Awesome work!
Is this PR ready to be reviewed or are you still working on it?

MartiDvrk:
> Awesome work! Is this PR ready to be reviewed or are you still working on it?

I think now it should be ready for review :)

nkrah:
Perfect. Give me a couple of days to go through the code, please. It's quite a bit. ;-)

andiresch:
There are two general aspects to be highlighted:
- we added a int EndOfRunActionMasterThread() similar to the corresponding Begin function that already existed. However, this function returns an int which is used to exit the remaining list of runs (and directly go to the last run - last run is needed to trigger EndOfEventWorkerAction). Note, this is slightly inconsistent, all other functions are void. 
- as discussed with David, we converted the 4D ITK dose images to "3D" thread local dose maps. As I didn't manage to use ITK images in the thread local data structure, a std::vector was used instead referred to as flat image (this was not discussed and may not be the smartest way to do it). Yet, we still use the ITK functionalities as before to not reinvent the wheel.

Feature request to future us: get rid of the mutex in the stepping action, which is still needed for unknown reason.

PR #300, opened 2023-11-20
Source activity 2

majacquet:
do not merge, just for wheels


tbaudier:
See https://github.com/OpenGATE/opengate/pull/287 for the first commits

PR #316, opened 2023-12-07
Switch to simpleitk

nkrah:
First steps to try the switch from itk to SimpleITK which is much faster to import. 

The general idea is to implement wrapper functions in opengate.image, e.g. itk_get_image_origin(image), which internally call the library specific functions to get the return value. 
Anything in opengate outside of the image module should exclusively call functions implemented in the image module, and NOT itk or SimpleITK directly. In this way, we can switch from one image library to another quite easily. 

The PR is a draft for now, but I am filing it as PR so everyone can have a look.


dsarrut:
Update
- yes indeed, ok for such wrapper functions
- SimpleITK may not be required: it seems that it may be linked to torch + itk together
- alternative idea: compile our own itk module with only the limited set of ImageType

nkrah:
Thanks. Let's keep the PR simmering for now and discuss how to move on.

dsarrut:
See 
- https://discourse.itk.org/t/torch-import-time/6354/3
- https://github.com/InsightSoftwareConsortium/ITK/pull/4369

nkrah:
Awesome, Thomas! So we can just keep the itk package then. 

I will go through this PR here and see which of the functions I have implemented might still be useful. For example, getting the size of an ITK image requires multiple lines and it is nice to provide the GATE user with a shortcut function. 

I keep the PR in Draft for now until I brush it up.

PR #318, opened 2023-12-08
Test simu dicom rt plan

majacquet:
Here is my simulation, designed to interpret a DICOM RT plan and execute a photon beam irradiation on a patient in accordance with the specified plan. 

To validate the functionality, I randomly selected four Control Points from the RT-plan. Subsequently, I computed the theoretical area generated by the collimation system aperture. To validate the built simulation, I simulated an alpha source at 1 MeV in a vacuum, determining its passage (or obstruction) through the collimation system before its interception in the first water voxel encountered within the simulated water tank. This simulation enables a direct comparison between the theoretically calculated area and the simulated area, derived from the summation of voxel areas triggered by the alpha beam. A predefined tolerance of 8% has been established to gauge the similarity between the two resulting areas.


tbaudier:
For the moment the test069 does not work

PR #322, opened 2023-12-13
Compton and Brem splitting

majacquet:
Add of the G4 biasing operator with the form of an actor to apply to a specific volume. Therefore, creation of an test operator which "actorify" the GEANT4 example proposing the Brem splitting.

Basically : Creation of an actor (BremSplittingActor) which inherit from two classes (GateVActor and G4VOperator). This actor call "operations" (here juste one operation :  the brem splitting process)  which is performed for each track on the desired volume.

At the same time, the physics list has to be modified (call of the special constructor class GenericBiasingPhysics) to include biased particles and biased processes applied to each of these particles. 

PS : sry for the name of the branch which should be brem splitting and not Compton splitting.


majacquet:
Update of the pull request. I created an other actor (with the test 071) which generates a biasing on the compton scattering, splitting it N times.

Basically, the actor, linked to a specific volume, is triggered each time a compton process ocurred. When a compton process ocurres, it will call N times the process compton, to generate the corresponding particles (gamma + electron)

The tests is basically the same as the 70, except the validation test is a comparison of each  parameter of interest between :
- A ref dataset generated without biasing physics
- The current dataset generated with the biased physics

opengate/tests/src/test071_operator_russian_roulette.py
```diff
@@ -0,0 +1,207 @@
...
+    s.track_types_flag = True
+    sim.physics_manager.physics_list_name = "G4EmStandardPhysics_option2"
+    ## Perhaps avoid the user to call the below boolean function ? ###
+    sim.physics_manager.special_physics_constructors.G4GenericBiasingPhysics = True
+    sim.physics_manager.processes_to_bias.gamma = list_processes_to_bias
+    #### Extremely important, it seems that GEANT4, for almost all physics lists, encompass all the photon processes in GammaGeneralProc
+    #### Therefore if we provide the name of the real process (here compt) without deactivating GammaGeneralProcess, it will not find the
+    #### process to bias and the biasing will fail
+    s = f"/process/em/UseGeneralProcess false"
+    sim.add_g4_command_before_init(s)
```
dsarrut, line 188: same for those lines, could be done auto

dsarrut:
- please add a short doc

PR #334, opened 2024-01-16
Added Modules for Optical Surface Creation 

guneet19:
- Made changes in managers.py to store surfaces info
- Made changes in engines.py to create optical surfaces
- Added pybinds required to create optical surfaces
- Added a testcase to verify the code changes

### DIFF
--- a/core/opengate_core/g4_bindings/pyG4OpticalSurface.cpp
+++ b/core/opengate_core/g4_bindings/pyG4OpticalSurface.cpp
@@ -1,0 +1,102 @@
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+#include "G4OpticalSurface.hh"
+#include "G4SurfaceProperty.hh"
+#include "G4MaterialPropertiesTable.hh"
+
+namespace py = pybind11;
+
+void init_G4OpticalSurface(py::module &m) {
+
+    // Binding for G4OpticalSurfaceModel enumeration 
+    py::enum_<G4OpticalSurfaceModel>(m, "G4OpticalSurfaceModel")
+        .value("glisur", G4OpticalSurfaceModel::glisur)  
+        .value("unified", G4OpticalSurfaceModel::unified)  
+        .value("LUT", G4OpticalSurfaceModel::LUT)  
+        .value("DAVIS", G4OpticalSurfaceModel::DAVIS)  
+        .value("dichroic", G4OpticalSurfaceModel::dichroic)  
+        .export_values();
+
+    // Binding for G4OpticalSurfaceFinish enumeration
+    py::enum_<G4OpticalSurfaceFinish>(m, "G4OpticalSurfaceFinish")
+        .value("polished", G4OpticalSurfaceFinish::polished)
+        .value("polishedfrontpainted", G4OpticalSurfaceFinish::polishedfrontpainted)
+        .value("polishedbackpainted", G4OpticalSurfaceFinish::polishedbackpainted)
+        .value("ground", G4OpticalSurfaceFinish::ground)
+        .value("groundfrontpainted", G4OpticalSurfaceFinish::groundfrontpainted)
+        .value("groundbackpainted", G4OpticalSurfaceFinish::groundbackpainted)
+        .value("polishedlumirrorair", G4OpticalSurfaceFinish::polishedlumirrorair)
+        .value("polishedlumirrorglue", G4OpticalSurfaceFinish::polishedlumirrorglue)
+        .value("polishedair", G4OpticalSurfaceFinish::polishedair)
+        .value("polishedteflonair", G4OpticalSurfaceFinish::polishedteflonair)
+        .value("polishedtioair", G4OpticalSurfaceFinish::polishedtioair)
+        .value("polishedtyvekair", G4OpticalSurfaceFinish::polishedtyvekair)
+        .value("polishedvm2000air", G4OpticalSurfaceFinish::polishedvm2000air)
+        .value("polishedvm2000glue", G4OpticalSurfaceFinish::polishedvm2000glue)
+        .value("etchedlumirrorair", G4OpticalSurfaceFinish::etchedlumirrorair)
+        .value("etchedlumirrorglue", G4OpticalSurfaceFinish::etchedlumirrorglue)
+        .value("etchedair", G4OpticalSurfaceFinish::etchedair)
+        .value("etchedteflonair", G4OpticalSurfaceFinish::etchedteflonair)
+        .value("etchedtioair", G4OpticalSurfaceFinish::etchedtioair)
+        .value("etchedtyvekair", G4OpticalSurfaceFinish::etchedtyvekair)
+        .value("etchedvm2000air", G4OpticalSurfaceFinish::etchedvm2000air)
+        .value("etchedvm2000glue", G4OpticalSurfaceFinish::etchedvm2000glue)
+        .value("groundlumirrorair", G4OpticalSurfaceFinish::groundlumirrorair)
+        .value("groundlumirrorglue", G4OpticalSurfaceFinish::groundlumirrorglue)
+        .value("groundair", G4OpticalSurfaceFinish::groundair)
+        .value("groundteflonair", G4OpticalSurfaceFinish::groundteflonair)
+        .value("groundtioair", G4OpticalSurfaceFinish::groundtioair)
+        .value("groundtyvekair", G4OpticalSurfaceFinish::groundtyvekair)
+        .value("groundvm2000air", G4OpticalSurfaceFinish::groundvm2000air)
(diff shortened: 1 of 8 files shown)


opengate/engines.py
```diff
@@ -291,6 +368,68 @@ def create_g4_optical_properties_table(material_properties_dictionary):
...
+    g4_surface = g4.G4OpticalSurface(g4.G4String(surface_name))
+
+    # Set Model
+    model = surface_properties_table["base_properties"]["surface_model"]
+    model_enum = getattr(g4.G4OpticalSurfaceModel, model, None)
+
+    if model_enum is not None:
+        g4_surface.SetModel(model_enum)
+    else:
+        fatal("Model in not present in SurfaceProperties.xml")
```
nkrah, line 385: Better: f"Unknown model {model}." 
Potentially add: 
"Available models are: g4.G4OpticalSurfaceModel.__members__.values()"
(not tested)

opengate/engines.py
```diff
@@ -291,6 +368,68 @@ def create_g4_optical_properties_table(material_properties_dictionary):
...
+
+    # Set Type
+    surface_type = surface_properties_table["base_properties"]["surface_type"]
```
nkrah, line 388: as above: I would prefer surface_type_name and surface_type
The _enum suffix is misleading because it suggests it's an enum, while it is just extracted from an enum.

opengate/engines.py
```diff
@@ -291,6 +368,68 @@ def create_g4_optical_properties_table(material_properties_dictionary):
...
+    surface_type_enum = getattr(g4.G4SurfaceType, surface_type, None)
+
+    if surface_type_enum is not None:
+        g4_surface.SetType(surface_type_enum)
+    else:
+        fatal("Surface Type is not present in Geant4 database")
```
nkrah, line 394: f"Unknown surface type {surface_type_name}"

opengate/engines.py
```diff
@@ -315,6 +454,10 @@ def __init__(self, simulation_engine):
         self.g4_em_parameters = None
         self.g4_parallel_world_physics = []
         self.g4_optical_material_tables = {}
+        self.g4_physical_volumes = []
```
nkrah, line 457: Please do not store extra references to the G4PhysicalVolumes here. They are accessible via every volume object as volume.g4_physical_volumes.

opengate/engines.py
```diff
@@ -499,6 +647,64 @@ def initialize_optical_material_properties(self):
...
+        )
+
+        for key, surfaces in volume_surfaces_info.items():
+            for surface in surfaces:
+                volume_1 = surface["volumes"][0]
+                volume_2 = surface["volumes"][1]
+
+                # Calls the function to get the physical volumes
+                # required to create optical surface (G4LogicalBorderSurface)
+                self.g4_physical_volumes = get_g4_physical_volumes(volume_1, volume_2)
```
nkrah, line 670: Better: get the GATE volume objects: 
vol1 = self.simulation_engine.get_volume(surface["volumes"][0])
vol1.g4_physical_volumes ...

opengate/engines.py
```diff
@@ -499,6 +647,64 @@ def initialize_optical_material_properties(self):
...
+
+                # Creates a surface properties table with information from xml.
+                self.g4_surface_properties = create_g4_surface_properties(
+                    surface_properties, surface["surface_name"]
+                )
+
+                print("Entering this initialize_surface_material_properties block")
+
+                # Creates an Optical Surface with the surface finish specified by the user
+                self.g4_logical_surface = g4.G4LogicalBorderSurface(
```
nkrah, line 686: g4_logical_surface should be a local variable because it changes at every iteration of this loop.

opengate/engines.py
```diff
@@ -499,6 +647,64 @@ def initialize_optical_material_properties(self):
...
+                    g4.G4String(surface["surface_name"]),
+                    self.g4_physical_volumes[0],
+                    self.g4_physical_volumes[1],
+                    self.g4_surface_properties,
+                )
+
+                print(f"The surface created is {self.g4_logical_surface}")
+
+                if self.g4_logical_surface is not None:
+                    self.g4_logical_surface_list.append(self.g4_logical_surface)
```
nkrah, line 696: I would prefer a dictionary self.g4_logical_surfaces rather than a list. The keys could be the same as in volume_surfaces_info (in that case, you do not to iterate over .items() - see above)

opengate/managers.py
```diff
@@ -452,6 +452,14 @@ class PhysicsManager(GateObject):
                 "is_input_file": True,
             },
         ),
+        "surface_properties_file": (
+            Path(os.path.dirname(__file__)) / "data" / "SurfaceProperties.xml",
+            {
+                "doc": "Path to the xml file containing the surface material properties to be used by G4LogicalBorderSurface. "
+                "Default: file shipped with Gate.",
```
nkrah, line 459: f"Default shipped with GATE located in {Path(os.path.dirname(__file__)) / "data" / "SurfaceProperties.xml"}"

opengate/managers.py
```diff
@@ -577,6 +589,51 @@ def dump_production_cuts(self):
...
+        volume_2 : The volume to.
+        surface_name : The name of the finish to be applied to the surface.
+
+        Returns:
+        None
+        """
+
+        surface_info = {"volumes": [volume_1, volume_2], "surface_name": surface_name}
+
+        volume_list = self.simulation.volume_manager.volumes
```
nkrah, line 607: This is a dictionary, not a list.

opengate/managers.py
```diff
@@ -577,6 +589,51 @@ def dump_production_cuts(self):
...
+
+        # Checks if the surface info of this volume pair already exists
+        # If exists, append. Else, create a new key pair
+        if volume_1 in volume_list and volume_2 in volume_list:
```
nkrah, line 611: Use sets to check this first, e.g.
```python
if not {volume_1, volume_2}.issubset(elf.simulation.volume_manager.volumes): 
    fatal(f"Unknown volume(s): {volume_1} and/or {volume_2}. "
```

opengate/managers.py
```diff
@@ -577,6 +589,51 @@ def dump_production_cuts(self):
...
+            if volume_1 in self.volume_surfaces_info:
+                self.volume_surfaces_info[volume_1].append(surface_info)
```
nkrah, line 613: I suggest creating a unique key as tuple of names: 
```python
unique_key = (volume_1, volume_2)
```
and use that as dictionary key. In this way, there is no need to a list of surface info under the same common volume name.

opengate/engines.py
```diff
@@ -499,6 +647,64 @@ def initialize_optical_material_properties(self):
...
+        information required to create optical surfaces.
+
+        This information stored is used to create an optical surface (G4LogicalBorderSurface).
+        """
+
+        volume_surfaces_info = (
+            self.simulation_engine.simulation.physics_manager.volume_surfaces_info
+        )
+
+        for key, surfaces in volume_surfaces_info.items():
```
nkrah, line 663: The variable key seems unused. Just iterate over volume_surfaces_info.values()

opengate/managers.py
```diff
@@ -577,6 +589,51 @@ def dump_production_cuts(self):
...
+        else:
+            # If either volume_1 or volume_2 is not present, raise a fatal error
+            missing_volumes = [
+                vol for vol in [volume_1, volume_2] if vol not in volume_list
+            ]
+            fatal(
+                f"Volume(s) {', '.join(missing_volumes)} not present in created volumes"
+            )
+
+    def dump_surface_information(self):
```
nkrah, line 625: If we implement a class 
```python
class OpticalSurface(GateObject): 
...
```
with volume_1, volume_2, surface_name as user_info, we get a dump_info for free from the base class. 

(See general comment)

nkrah (commented):
Looks great. Some changes here and there will make the code smoother. 
A general change that I suggest is to implement a python class OpticalSurface inheriting from GateObject. The relevant info, i.e. volume_1, volume_2, surface_name, can be configured and stored as user_info. The class provides its own initialize() method. The PhysicsManager keeps a dictionary of OpticalSurface objects and the PhysicsEngine iterates over them to initialize each. The Region class behaves like this - you might have a look. 
The advantage is that we get a lot of functionality for free via the base class GateObject and JSON dump will work out-of-the-box. 

We can discuss via Zoom how to implement the class, but I think it requires not many changes, but would improve the code structurally.

core/opengate_core/g4_bindings/pyG4OpticalSurface.cpp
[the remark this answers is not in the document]
guneet19: Checked this by removing these statements. It did not throw any segmentation fault.

dsarrut:
merged with #388

PR #354, opened 2024-02-19
Compton splitting + pseudo transportation

majacquet:
Do not merge for the moment.

Here is the basement for a Variance reduction method implying photon pseudo-transportation through a specific media. When a photon is tracked inside a media and undergoes a Compton process, the number of particle generated by this process are split N times. All the split particle (except the primary one) are transported through the biased volume without any interactions. A weight according to the splitting value and the probability to not interact in the volume is set. The primary particle still undergoes normal interaction, but is killed if it exits the biased volume.


majacquet:
The test 072 verifies if the transportation is well realized. When a photon undergoes a Compton interaction, we retrieve the vertex of the pseudo-transported particle, the position of detection , the kinetic energy of the particle and its weight. With all of these information, we can calculate the attenuation coefficient which is then compared to the one calculated from NIST-XCOM.

dsarrut:
(@majacquet) do not merge. Keep the PR for later

PR #355, opened 2024-02-20
first tentative coinc sorter

dsarrut:
Warning uproot NEED to be updated to 5.2.2 

First tentative
- the file 'coincidences.py' will contain all the algorithms
- the tests 72 do not test anything yet, it is just a starting point


dsarrut:
See also #372

dsarrut:
See #402

PR #369, opened 2024-03-12
Source with angles as histogram

garaymond:


tbaudier:
@garaymond 
The test test010_generic_source_ene_ang.py is not working. I let you modify it. Thx

PR #377, opened 2024-03-20
Simple geometrical splitting actor

majacquet:
Add an actor which duplicates a track entering and/or exiting the actor volume.


dsarrut:
(@majacquet) do not merge, but keep the PR for the moment

PR #385, opened 2024-04-09
Back to back source

dsarrut:
Based on #381.
@MaxTousss : I duplicate and consolidate this PR to make it standalone.
If you agree, it can be merged in the master while you can still continue your branch (#381)

### DIFF
--- a/core/opengate_core/opengate_lib/GateGenericSource.cpp
+++ b/core/opengate_core/opengate_lib/GateGenericSource.cpp
@@ -13,12 +13,6 @@
 #include <G4UnitsTable.hh>
 
 GateGenericSource::GateGenericSource() : GateVSource() {
-  /*fNumberOfGeneratedEvents = 0;
-  fMaxN = 0;
-  fActivity = 0;
-  fHalfLife = -1;
-  fLambda = -1;
-  */
   fInitGenericIon = false;
   fA = 0;
   fZ = 0;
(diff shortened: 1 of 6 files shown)


dsarrut:
sorry for the confusion ;)
- I think a simple "back to back" source is interesting for some users, without waiting for acoli stuff
- I started this new PR to merge it independently while still allowing you to continue #381 at your pace 

Maybe I overlooked it, I thought test072 was really testing something, this is not the case? 

The list of attributes is not really comprehensively described, there is only a rough documentation here : 
https://opengate-python.readthedocs.io/en/latest/user_guide.html#digitizerhitscollectionactor
Position, Direction refer to the step. Pre/Post refer to pre step and post step (see [geant4 doc](https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/html/TrackingAndPhysics/tracking.html#access-to-track-and-step-information)). When "Event" is prepended, it refers to the Event (primary particle) instead of the step.

hope that helps!

MaxTousss:
That explains the "merge now"!

As for the "really testing something", yes it does but:
- It was "broken" for debug before [e8dfcef](https://github.com/OpenGATE/opengate/pull/385/commits/e8dfcefbd7a9936d907c2309a33e4e47cc0155dc) commit. Now that you modified that, it should behave as intended.
- The tests were made with my limited knowledge of attributes and brute-force, as shown by the hacky way I uses to find true back-to-back (`b2b_dir`). If you don't mind that, then yes, it is testing things. However, I do not claim that the tests are sufficient. For example, I still doesn't understand why e+ and e- particles seems to be born from gamma photons interacting with the air/void...

dsarrut:
OK, so I merge. 
You can continue #381 to bring a little bit of accolinearity to those boring perfectly back to back gammas ;)

PR #401, opened 2024-05-07
Dose actor secondary parts

majacquet:
Add of a new dose actor (only monothread) to only deposit the secondary dose generated by photons interacting on matters. (This actor will not work for the moment above ~ 8 MeV primary photons  because of neutron generation)


nkrah:
Great! Please do not merge before actor refactoring is done. 
We will need to restructure this new dose actor. Let's discuss how this can best be achieved.

nkrah:
@majacquet : What's the status of this PR? Still of interest? It will require adaptation to the new style actors.

PR #403, opened 2024-05-13
Refactor actors - towards merge

nkrah:
created based on PR #363 as a branch in the main repo. 

Status : https://github.com/OpenGATE/opengate/wiki/Status-of-test-for-refactor%E2%80%90actors-branch


nkrah:
CHECK that commit
https://github.com/OpenGATE/opengate/commit/bd743c9564379a1356640f36ff0c4a9c56e075ea
is considered in this PR!

dsarrut:
Status:
- WARNING: I change all output files to the output folder (NOT in the current folder), this is important for users using opengate_tests 
- tests 075, 074, 073, 072 = OK
- test 071 and 070 needs debug (Nils)
- test 068: OK
- test 067: run but fails. To check

nkrah:
* Fixed test073 (there were still problems)
* Fixed TesselatedSolid and worked on test067_stl... -> works intermittently. Sometimes tests pass, sometimes not. Need to understand better. 
* I also get occasional segfaults, but they do not seem to be related to the specifics of the test, rather related to nuclear physics components. Maybe some change in G4? (I was still on the previous version until very recently)

nkrah:
General comment: 
I strongly advise to set the output directory via Simulation.output_dir and define the output_filename via the actor. GATE will automatically combine the two. This is much easier to maintain than concatenating the folder and the filename in the output_filename of each actor. If the user wants to change the output location, she just needs to change the single parameter Simulation.output_dir. 
... also more robust to maintain in the tests here ...

dsarrut:
Status (starting from the last tests): 
 
- 072 to 075 : OK
- test071_test_operator_compt_splitting.py ==> NILS
- test070_test_operator_brem_splitting.py ==> NILS
- 068: OK
- test067_stl_volume.py ==> unknown seg fault bug
- 066 : OK
- test065_dose2water_from_edep2water_ct.py test065_dose_from_edep_ct.py ==> BUG 
   VERY SLOW between "initialize G4RunManager" and "initialize PhysicsEngine"
- test065_dose_from_edep_volume.py = OK
- test065_sim_as_dict.py = OK
- test065_sim_from_dict.py = OK
- 064: OK
- 063: OK
- 062: OK
- 061: OK
- 060: OK

nkrah:
Regarding test065_dose_from_edep_ct.py: 
I added a timing test of the method `patient.create_label_image()`. Creating the label image (voxel labels) takes about 5.5 sec on my MacBook Pro. 

How "very slow" is the image read on your machine?

opengate/actors/filters.py
```diff
@@ -23,9 +23,21 @@ class FilterBase(GateObject):
...
 
+    def __initcpp__(self):
+        """Nothing to do in the base class."""
+
     def initialize(self):
         self.InitializeUserInput(self.user_info)
 
+    def __getstate__(self):
+        state_dict = super().__getstate__()
+        state_dict["filter_engine"] = None
```
nkrah, line 34: Why do we need to set this to None? I have the impression that the filter_engine attribute is actually never set, so there should not be any need for setting it to None. I think the __getstate__ method is not needed. The __setstate__ method yes.

dsarrut:
> Regarding test065_dose_from_edep_ct.py: I added a timing test of the method `patient.create_label_image()`. Creating the label image (voxel labels) takes about 5.5 sec on my MacBook Pro.
> 
> How "very slow" is the image read on your machine?

this is (still) due to torch inside itk. If you dont have torch installed, it is fast. I will try to investigate, but we may consider to 
switch to SimpleITK (much faster)

nkrah:
I see.
  
​Indeed the disadvantage of the LazyLoaeder mechanism is that the imports happen at some point during run time. Makes debugging slow code more difficult. Something to consider.
  
​Yes, simpleITK could be an option, but that is definitely an extra PR with quite a bit of work because ITK is scattered in many places.

nkrah:
Quick update on the DoseActor segfault: After cleaning the GateDoseActor (C++) class, the segfault only appears when activating any squared scoring in the dose actor. Scoring squared quantities implies instantiating a thread local struct and I therefore believe that the segfault is somehow caused by that mechanism. Will further look into it.

dsarrut:
WARNING to #[34ccbce](https://github.com/OpenGATE/opengate/pull/403/commits/34ccbce7dc26f2c4e9ed7f5e841ca93b40997efb) 
I change the way image data are sent from py to cpp. Now this is explicitely copied on the cpp side. I had to do this to avoid (all? some?) seg fault. Before that it works (sometimes) on osx, but always seg fault on linux.

To be discussed ...

nkrah:
Intéressant ça. On en discute.
  
  
  
  
  
  
  

  
  
>

dsarrut:
https://github.com/OpenGATE/opengate/wiki/Status-of-test-for-refactor%E2%80%90actors-branch

nkrah:
Started implement a mechanism to have interfaces to the user_output in actors. The idea is that the user accesses the user_output via these interfaces to hide the underlying complexity. 
Important: The mechanism works, but I have not yet finished cleaning up. Except bugs introduced by these recent changes. I'll fix them this weekend.

dsarrut:
hurray !!

Issue #412, opened 2024-06-10
Change force_rotation option


dsarrut:
https://github.com/OpenGATE/opengate/blob/master/opengate/tests/src/test068_rotation_source.py
When we want the source direction to follow the orientation of the attached volume, the option is 'force_rotation'. 
To be changed to "direction_relative_to_attached_volume"

Also, change the elektaversa e- source example.

dsarrut:
See #424

PR #416, opened 2024-06-13
Generic source : direction relative to attached volume

dsarrut:
- replace the flag force_rotation #412 
- apply direction rotation also for isotropic direction + restricted angle
- no test yet !


dsarrut:
Done in #419

PR #419, opened 2024-06-14
Generic source : direction relative to attached volume

axelrannou:
Update test068 : Add an isotropic source
No need for test075


dsarrut:
Thanks a lot, @axelrannou, can you please add some rotation to the plane to check that the directions are correctly modified?

axelrannou:
> Thanks a lot, @axelrannou, can you please add some rotation to the plane to check that the directions are correctly modified?

Hi, isn't it that already done with [motion_source actor](https://github.com/OpenGATE/opengate/blob/079f16bf6737dad4ff030b6735ba36f3456a936e/opengate/tests/src/test068_rotation_source.py#L125-L126) and [rotation](https://github.com/OpenGATE/opengate/blob/079f16bf6737dad4ff030b6735ba36f3456a936e/opengate/tests/src/test068_rotation_source.py#L155) ?

dsarrut:
exact ! No  need so, thanks !

PR #422, opened 2024-06-28
C-arm contrib

axelrannou:
Add a model that simulates the use of a c-arm (siemens cios alpha)
- rotation around the center of the c-arm
- collimation
- test TODO


dsarrut:
very good, this will be the first CT example ;)

dsarrut:
@tbaudier  it should not be closed yet : the test 075 do not really test.

PR #439, opened 2024-08-22
set SetMeanEnergyPerIonPair for acollinearity

dsarrut:
@MaxTousss for you ;)

### DIFF
--- a/core/opengate_core/g4_bindings/pyG4IonisParamMat.cpp
+++ b/core/opengate_core/g4_bindings/pyG4IonisParamMat.cpp
@@ -13,6 +13,9 @@ namespace py = pybind11;
 
 void init_G4IonisParamMat(py::module &m) {
   py::class_<G4IonisParamMat>(m, "G4IonisParamMat")
-      .def("GetMeanExcitationEnergy",
-           &G4IonisParamMat::GetMeanExcitationEnergy);
+      .def("GetMeanExcitationEnergy", &G4IonisParamMat::GetMeanExcitationEnergy)
+      .def("GetMeanEnergyPerIonPair", &G4IonisParamMat::GetMeanEnergyPerIonPair)
+      .def("SetMeanExcitationEnergy", &G4IonisParamMat::SetMeanExcitationEnergy)
+      .def("SetMeanEnergyPerIonPair",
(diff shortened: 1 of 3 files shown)


MaxTousss:
Hi @dsarrut  

So, I am going to make some time this week to try and push acolin for back-to-back. Can you remind me what you wanted me to do with this PR? It seems you already have the code and some tests. Do you want me to code review or is there something missing for this PR?

dsarrut:
thanks @MaxTousss ! Yes if you can check that everything is fine, it would be nice. 
If I remind correctly:
- this test is to illustrate G4 accol is working with ion source, provided the option (SetMeanXXX) is set. Does it seems ok for you? 
- now another test + code modification is expected for the back to back, that should provide similar results with faster source, am I right ? 

If it is fine for you, can you also add some word in the documentation (and advertise for your publication) please ? 
thanks !

MaxTousss:
The code seems fine and the test result fit with what I had done with Gate 9.X. However, I have three questions:
- Have you asked/validated that setting a mean energy to the materials does not affect anything else? That was one of the main reason why I did not push to merge this modification to Gate 9.X. For example, could it have an impact on positron range? 
- From looking at the test, it seems that you intend for the user to set the value "manually" in the sense that Gate would not, by default, enable accolinearity for Ion source. This is consistent with Gate prior to version 10, except that here (v10) the user modify the materials property vs the physics (before v10). Just to make sure, is that intent voluntary? 

Also, do you mind and can you enable me to make some changes to the test file? I want to add a Rayleigh fit so that the figure produced by the test talk by itself. I would also add the documentation in this PR.

dsarrut:
- no, I did not check and I dont know if it can have an impact. Ideally, at test (positron range) could be done to verify. But maybe, first a note in the document is sufficient
- yes, I usually prefer that this is explicit. Indeed, this is not fully clear yet (user set `SetMeanEnergyPerIonPair` not something like `enable_accolinearity`) but lets say it is a first step. We can still add more user friendly command later what we have a bit more experience
- please do every change you want 😄 !

MaxTousss:
I do not have write access on this branch/PR... So, I guess I will branch from this an create a new PR, saying to close this one when/if the new PR is merged?

MaxTousss:
This PR is currently being continued in #485

Issue #338, opened 2024-01-23
Qt6 support


BishopWolf:
Geant4 is moving to qt6 (slowly but moving: https://geant4.web.cern.ch/download/release-notes/notes-v11.2.0.html).

Therefore, Gate should be able to compile in qt6. This means gate_core built against qt6 and all pyqt calls using pyqt6. 
Which version? Obviously the one supported by geant4 but 6.6 should be fine.

dsarrut:
Yes ! We will update the current version to Geant4 11.2 in the following months and look at qt6 at that moment. 
Currently, the qt support within opengate is not working well. Maybe it will be better with qt6 !

pavelToman:
Hello @tbaudier,
thank you for a quick reply!

My sources are for v10.0.2 from https://github.com/OpenGATE/opengate/archive/refs/tags/10.0.2.tar.gz
So I will fetch the master instead :)

PR #442, opened 2024-08-26
Use qt6 for mac instead qt5

tbaudier:


BishopWolf:
You shall move completely to QT6 in all systems. QT5 is mostly in maintenance status and will die eventually. Moreover, QT6 solves a lot of issues, specifically for PyQt.

nkrah:
As a side note: I have been experiencing compile issues on my Mac with Qt5 (installed via brew) and geant4 recently. Ended up compiling geant4 without Qt for now. In any case a good idea to move to Qt6

BishopWolf:
Linked issue https://github.com/OpenGATE/opengate/issues/338

BishopWolf:
"[Remove macos14 because python does not work now]"

This is because python 3.8 doesn't work on mac14, you shall drop python 3.8 support for that.

tbaudier:
> "[Remove macos14 because python does not work now]"
> 
> This is because python 3.8 doesn't work on mac14, you shall drop python 3.8 support for that. Most systems (including debian 11 Bullseye) are on minimum python 3.9 anyway.

Here, the error was due to that problem: https://github.com/actions/setup-python/issues/825
BTW, 3.8 will be drop

BishopWolf:
Beware of this warning (it should be an error):
```
/Users/runner/miniconda3/envs/opengate_core/lib/python3.10/site-packages/setuptools/command/build_py.py:215: _Warning: Package 'opengate_core.plugins' is absent from the `packages` configuration.
```

tbaudier:
> Beware of this warning (it should be an error):
> 
> ```
> /Users/runner/miniconda3/envs/opengate_core/lib/python3.10/site-packages/setuptools/command/build_py.py:215: _Warning: Package 'opengate_core.plugins' is absent from the `packages` configuration.
> ```

'opengate_core.plugins' contains the qt librairies needed by QtCore to display. You can find libqcocoa for exemple. No python scripts are available in that folder so I'm not sure the folder should be in the packages configuration.

I need to create that folder because delocate could not find theses librairies by itself

BishopWolf:
I see you are testing brew instead of conda, just a hint: you can't have mixed packages with BOTH installers. Therefore, to use brew you must drop conda completely

tbaudier:
> I see you are testing brew instead of conda, just a hint: you can't have mixed packages with BOTH installers. Therefore, to use brew you must drop conda completely (for all packages) watch this line
> 
> https://github.com/OpenGATE/opengate/blob/20b44ad0c91b617a7b61dee132b3076f0c29cfe3/.github/workflows/main.yml#L107

conda is just used to install qt6 c++ librairies. I do not use conda to install python packages. For macOS, I use Github Actions for intel Mac and Cirrus for M1. On M1 brew do not succeed to install qt6 correctly, this is why I used conda.

PR #450, opened 2024-09-05
[WIP] apply Snyk suggestions

BishopWolf:
Clean issues caught by Snyk

### DIFF
--- a/.github/workflows/delocateWindows.py
+++ b/.github/workflows/delocateWindows.py
@@ -48,7 +48,7 @@ def hash_filename(filepath, blocksize=65536):
 
 def find_dll_dependencies(dll_filepath, lib_dir):
     print(dll_filepath)
-    if dll_filepath in global_dll_deps.keys():
+    if dll_filepath in global_dll_deps:
         return global_dll_deps[dll_filepath]
     else:
         dll_deps = {}
--- a/opengate/actors/doseactors.py
+++ b/opengate/actors/doseactors.py
@@ -49,6 +49,7 @@ class DoseActor(g4.GateDoseActor, ActorBase):
 
     type_name = "DoseActor"
 
+    @staticmethod
     def set_default_user_info(user_info):
         ActorBase.set_default_user_info(user_info)
         # required user info, default values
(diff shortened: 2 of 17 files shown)


BishopWolf:
Some issues description
1. use of not element in list: this must be replaced by element not in list. Explanation: the not is negating the element, you intend to negate the existence in the list, it is not the same.
2. use of f strings without replacements. f-strings always calls to replacements even if there is none, this creates unnecessary overload for simple strings.
3. use of file open outside with statements. This is highly dangerous, it may lead to broken files.
4. comparing to None. None is an Object and therefore you shall check using the is operator.
5. missing staticmethods.
6. use of exit. This is dangerous, we must use explicitly sys.exit
7. use of preset words (sum, slice, list) as variable names.

BishopWolf:
The issues are easy to fix, so you can incorporate them into the same PR. Just review the list of descriptions.

nkrah:
Thanks for the suggestions. We have indeed already incorporated many of them while doing the massive refactoring, e.g. opening files within a context, etc. We'll further consider your suggestions, possibly after finishing the current refactoring.

nkrah:
Concerning the exit(): We are planning to raise exceptions rather than calling sys.exit() wherever possible. What is your opinion on this?
(There are many "legacy" pieces of code that will be updated in further upcoming refactoring steps. )

nkrah:
Just went through your suggestions and implemented (in PR #403 404) those that were not anyhow already covered by PR #403 404. 

We'll have another look once the PR is done and merged.

BishopWolf:
> Concerning the exit(): We are planning to raise exceptions rather than calling sys.exit() wherever possible. What is your opinion on this? (There are many "legacy" pieces of code that will be updated in further upcoming refactoring steps. )

Raising/Handling exceptions is always preferable to exit.
You must add an GateException class to be called by everything.

Rules:
1. No exception remains unhandled
2. No generic exceptions
3. never leave a method silently

nkrah:
To follow up on this: most of the changes in this PR have been either addressed or the code has changed through other PRs. I don't think it makes sense to keep this PR open for merging. 

I guess the best would be to re-check with Snyk and start a PR from there. 

@BishopWolf : Ok for you? Would you be so kind and run Snyk over the current master and see what it reports?

BishopWolf:
> @BishopWolf : Ok for you? Would you be so kind and run Snyk over the current master and see what it reports?

Sure, close this PR and I rerun Snyk in current master

PR #453, opened 2024-09-16
Hints from IDE

dsarrut:
Add python hints that should helps IDE to find dynamic attributes. At least pycharm seems happy with that. 
Only for base and actors right now. To be completed.


dsarrut:
merged in #455

PR #460, opened 2024-09-20
Prepare qt6 for linux

tbaudier:
For linux we need to change the manylinux version:
 - manylinux2014 (= manylinux_2_17) : gcc version is too old to compile qt6
 - manylinux_2_24 cannot be used because apt do not delivered packages (gcc, ...) for this version
 - manylinux_2_28: qt6 can be installed manually. But this version of manylinux could lead to compatibility version for older computer like computer center using Centos8.

What is needed:
 
- [x]  Push the docker to docker hub (which version?)
- [x]  Create a wheel for visualisation and one for cluster?
- [x] Update doc (https://opengate-python.readthedocs.io/en/latest/developer_guide.html#step-1-geant4-and-qt)

### DIFF
--- a/.github/workflows/Dockerfile_opengate_core_arm64
+++ b/.github/workflows/Dockerfile_opengate_core_arm64
@@ -1,12 +1,15 @@
 #Docker for opengate_core
 #systemctl start docker
 #login: docker login
-#build: docker build -t tbaudier/opengate_core -f Dockerfile_opengate_core .
+#build: docker build -t tbaudier/opengate_core -f Dockerfile_opengate_core_arm64 .
 #push: docker push tbaudier/opengate_core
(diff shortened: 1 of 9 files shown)


BishopWolf:
Interesting to read https://github.com/pcolby/dokit/blob/main/.github/workflows/build.yaml

BishopWolf:
> * manylinux_2_28: qt6 can be installed manually. But this version of manylinux could lead to compatibility version for older computer like computer center using Centos8

This is solved by using Centos9 that is already compatible with qt6.4+. Centos8 has serious compatibility issues here.

dsarrut:
| | test wo G4 | compil | wheel |
|--|---------| --|-------------
| mac |  | OK | 
|linux | |  OK |

PR #463, opened 2024-09-24
Update user_guide_2_1_volumes.md specify voxel_materials range

jizhang02:
The voxel_material range is a half-closed interval, i.e. [ ).  Because I made a mistake in my dose simulation when I did not pay attention to the organ/tissue label to the corresponding voxel materials. For example, [-900, -100], "Lung", in this case, the voxels belong to -100 are not assigned to "Lung", its actual range is [-900, -99].


dsarrut:
your are right, thank you!

PR #467, opened 2024-09-28
Simple refactoring of sources to use GateObject

nkrah:
A first "shallow" refactoring of the sources. The sole purpose is to let them become GateObjects so sources can use all the cool features that GateObjects have. Any other structural changes can be done later. 

To start with, I explained chatGPT how to generate user_info_default dictionaries based on the attributes defined in set_default_user_info(). It works quite well and chatGPT was even able to generate some 'doc' entries based on comments in the code. Saved a lot of manual work. 

PR opened in as a draft.


dsarrut:
#550

PR #472, opened 2024-10-02
Multiprocessing for opengate_tests

andiresch:
- opengate_tests can run all tests in individual processes as lot of time is spent in i/o; 
- reduction of calculation time of about factor 4 on my workstation
- one can choose between "legacy" = previous implementation, single processing and multi processing
- some additional info printed 

- Few test cases fail in multiprocessing mode (e.g. test072_coinc_sorter_2keepAll.py test072_coinc_sorter_2removeMultiples.py
- Test cases failed in mp are then restarted in sp. If it succeeds there, we write success to the dashboard dict --> not sure if this is a good approach


dsarrut:
very good, thanks a lot Andreas ! 

Some remarks/questions : 
- run_previously_failed_jobs : unclear why it is needed ? 

- warning, some tests are dependent on the output of other tests that must be run before (test040_gan_phsp_pet_gan.py or test072_coinc_sorter_2keepAll.py). This is currently done with subprocess. In a MP mode we may imagine some conflict if the same test is run twice at the same time (not very probable, I know, but, be sure it will happens ;). Unclear how to deal with that

- get_imported_modules : is that needed ? Warning we implemented a LazyModuleLoader that only load some (large) module when it is needed (torch, etc). 

- I did not catch how the MP works: everything is run at the same time and we rely on the machine to dispatch to several cores ? Because the number of tests is (usually) >> than the number of cores, it may lead to issues. Anyhow, I will test to see !

andiresch:
thanks for your comments, David!
- run_previously_failed jobs is needed as some tasks fail in multiprocess - and I did not yet solve why; most of them involve root files - but actually it may be the dependency on previous file: e.g. test072_coinc_sorter_2keepAll is failing for me most of the times. Thanks for this input, didn't know!
- get_imported_modules: wanted to generalize the avoid modules, but exactly as you thought the LazyLoader gives problems - and then I was too lazy to continue... ; I will remove this functionality
- number_tests >> number_cores: indeed! as fas as I observed, multiprocess.Pool runs in parallel as many processes as logical CPU cores available on the node and keeps the remaining function calls in the queue

andiresch:
- changed tests which have a dependency on other previous tests: now those require a main function with argument dependency = 'test_depending_on.py' - opengate_test checks for this argument and runs all those tests first
- works for almost all test cases, but not yet for:
test049_pet_digit_blurring_v4.py : failed
test058_iec_six_spheres.py : failed
test028_ge_nm670_spect_3_proj.py : failed
test038_gan_phsp_spect_gan_ze.py : failed
test028_ge_nm670_spect_3_proj_blur.py : failed
test040_gan_phsp_pet_training_dataset.py : failed
test053_phid_05_it_ref_mt.py : failed
test038_gan_phsp_spect_gan_se.py : failed
test038_gan_phsp_spect_training_dataset_mt.py : failed
test015_iec_phantom_3.py : failed

MaxTousss:
Oops, I started to look into that yesterday since the tests were taking too long. I should have looked at the available PR before doing that XD. I am happy to see that it is progressing nicely!

Here are some comments:
- I believe that tests should be independent from each other. If B depends on A, one could modify A and break B without knowing it until it launches the tests. Thanks to andiresch PR, we would at least know that B depends on A but the inverse remains a problem. While I can guess that these tests were probably built with the idea of limited the number of lines in the test, I believe that making sure the tests are coherent and future proof is better than saving lines.
    - Furthermore, this means that ``random_tests`` was not reliable even before this PR.
- You could provide the user an option to set the number of cores used and pass it to Pool as ``Pool(processes=num_workers)``. Some of the tests are multithreaded, so having as many processes as core is counter-productive, and I like to keep some leeway to do something else when the tests will run.
- It seems that the output of the testing does not follow the test numbering. Is that a problem for the user/dev? I believe that it is not a problem since I care only about the failed vs success but maybe there is a hidden advantage?
    - If the order is not important, I would suggest that the order in which the tests are run should be randomized. Thus, we would have a better chance to catch errors resulting from interdependencies and the like. The order should probably be recorded somewhere for potential debugging related to the order.

andiresch:
thanks for your comments!
- added the option to limit the number of processes -> in next commit 
- inter-dependency of tests: you are right that A could break B without knowing, but somehow those tests are very related and I think is very unlikely that one modifies A without looking at B. "random_tests" was working before, because each of those tests with a dependency was running a subprocess(A) to create the needed input. So just the number of jobs was not the one you would be higher than expected (we talk about order of less than 10 tests). Now both should be under control - and now I would suggest to require an explicit statement in the main that there is a dependency. Actually, we could also put an argument in "A" that "B" follows ... if we really want to. not sure is needed.
- I have no opinion about the order, but when running as a pool, we require that they are independent from each other, so is more a matter of printing the outcome, no?

PR #474, opened 2024-10-03
fix include position in GateInfo.cpp

alexis-pereda:
There was an inclusion (`#include <QtGlobal>`) in a function in `GateInfo.cpp`.


dsarrut:
thanks !

PR #482, opened 2024-10-08
Only download the g4 data that are missing

MaxTousss:
Currently, even if only one g4 data folder is missing, all of them are re-downloaded. This PR change this behavior by keeping track of which g4 data is missing and only downloading them. The naming of the variables/methods were adapted to this new behavior.

This is a really small quality of life feature for dev. It is mostly for seldom contributors like me that also have slow internet. I guess we could say that limiting the number of downloads also save electricity and thus our carbon footprint? :P


tbaudier:
Hello,

Thank you very much, I merge

PR #486, opened 2024-10-09
Multiprocessing

nkrah:
Enable GATE 10 to split a simulation into multiple parallel processes. 
THIS IS WORK IN PROGRESS

First implemented items: 
* split run timing intervals
* adapt dynamic objects (run-based)
* spawn processes via Pool
* write output into a separate subfolder per process

Still missing: 
* merge actor output from different processes

### DIFF
--- a/opengate/managers.py
+++ b/opengate/managers.py
@@ -7,6 +7,7 @@ import shutil
 import os
 from pathlib import Path
 import weakref
+import multiprocessing
 
 import opengate_core as g4
 
@@ -1460,6 +1461,7 @@ class Simulation(GateObject):
         self._current_random_seed = None
 
         self.expected_number_of_events = None
+        self.mapping_run_timing_intervals = {}
 
     def __str__(self):
         s = (
@@ -1653,7 +1655,7 @@ class Simulation(GateObject):
     def multithreaded(self):
         return self.number_of_threads > 1 or self.force_multithread_mode
 
-    def _run_simulation_engine(self, start_new_process):
+    def _run_simulation_engine(self, start_new_process, process_index=None):
         """Method that creates a simulation engine in a context (with ...) and runs a simulation.
 
         Args:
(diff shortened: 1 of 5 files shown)


opengate/managers.py
```diff
@@ -1668,17 +1670,74 @@ def _run_simulation_engine(self, start_new_process):
         with SimulationEngine(self) as se:
             se.new_process = start_new_process
             se.init_only = self.init_only
+            se.process_index = process_index
             output = se.run_engine()
         return output
 
-    def run(self, start_new_process=False):
+    def generate_run_timing_interval_map(self, number_of_processes):
+        if number_of_processes % len(self.run_timing_intervals) != 0:
```
dsarrut, line 1678: why ? I thought we just divide ALL time_interval by the number_of_processes

nkrah:
Yes, but letting the user define the total number of processes rather than the process per run is more intuitive and will not require an API change if we implement a more advanced splitting scheme in the future. So I think it's better this way.

opengate/managers.py
nkrah: Yes, but letting the user define the total number of processes rather than the process per run is more intuitive and will not require an API change if we implement a more advanced splitting scheme in the future. So I think it's better this way.

nkrah:
I figured out a flexible mechanism to merge data back into one single actor output (if data is mergeable: true for images, not true yet for ROOT). 
We will need a new type of method, common to all actors, namely FinalizeSimulation(), to be triggered from the Simulation after all processes have finished. Writing the combined output (from the processes) to disk will be done in FinalizeSimulation(). The EndOfSimulation(), where writing currently takes place, is called inside the process and therefore before combining the output. We can also add an option to not store intermediate, i.e. per process, output on disk if not needed. For example: images are accessible directly via memory and can be merged that way. No need to access data from disk. 

Note: FinalizeSimulation() will not have access to engines because they do not exist any more outside of the subprocess.

nkrah:
New: 
The following actors now work in multiprocessing (local machine):

* SimulationStatisticsActor: data is merged in memory and accessible after the simulation; written to disk if requested

* Actors with ROOT output: root files (from subdirectories per process) are merged into new root file in main output folder structure. Event IDs are automatically incremented. RunIDs are recreated as per the original simulation. 

Works with test019_phsp_actor -> created a new variant of the test. 

Still need to create variants of other tests that use ROOT output to check.

BishopWolf:
@nkrah I think all actors shall have atomic variables, this way all actors will be thread safe by default, watch this library https://pypi.org/project/atomicx/ . It already implemented the atomic doubles on my suggestion

```python
from atomicx import AtomicFloat

# Create an atomic float with an initial value of 0.0
atom = AtomicFloat()
print(f"Initial Value: {atom.load()}")

# Perform atomic operations
atom.store(3.14)
value = atom.load()
print(f"Value: {value}")

# See docs for more operations
```

nkrah:
@BishopWolf Thanks for the suggestion. I think atomic doubles will be useful for certain parts of the actors. 

Bear in mind that this PR is about multiprocessing, i.e. running a (independent) simulation in a newly spawn process. There is no issue with shared memory handling in this case. 

Concerning multithreading: We are actually using the multithreading architecture from Geant4 which means that not every part of a simulation runs in separate threads, only certain methods. Therefore, only certain shared data structures, e.g. images into which all threads write, need to be thread safe. Currently, there is no python-side function that accesses shared data on a per-thread basis, only C++ functions. In case this changes in the future, I think the package you suggest could be a good option.

nkrah:
I will pick this up again once PR #599 is done.

PR #487, opened 2024-10-10
Multiprocess tests

andiresch:
- now able to run opengate_tests with multiprocessing: speedup on my workstation of factor 7 (tests 0 - 20) and factor 4 (tests 60+)
- now all tests should work, needed to fix some paths (multiple tests writing to the same file)
- thanks to @tbaudier for your feedback

- one open question: what is the current recommended G4 version? I tried with 11.1.1 and 11.1.2; are we already at 11.2 ?


tbaudier:
> * one open question: what is the current recommended G4 version? I tried with 11.1.1 and 11.1.2; are we already at 11.2 ?

The version used for the wheels is 11.2.1

PR #499, opened 2024-10-19
Prepare PR #478 OptiGAN for merge

nkrah:
Review and adapt the code to match the GATE 10 architecture. 

In particular, 
* simplify handling of output directories. 
* Let the user provide a phase space actor as input and pick up as much information from there. 
* alternatively, allow for stand-alone operation based on root_file_path


nkrah:
I will check for remaining commented code to be removed before merging.

PR #501, opened 2024-10-21
Stop simulation on user criteria

MartiDvrk:
The aim of this PR is for Actors to be able to communicate to the SourceManager to stop the current run and move to the next one.
- merged PR 451 -> actors can access source manager.
- source manager has now a fRunTerminationFlag. This flag can be modified by the actors calling the SetRunTerminationFlag function on the source manager.

On the DoseActor, the user can now set a target statistical uncertainty (uncertainty_goal on the python side) :
- recovered calculation of mean uncertainty from before the refactoring of the actors.
- the mean uncertainty is evaluated by a single thread on the EndOfEventAction, only after a certain number of events, and compared to the goal uncertainty. When the goal is reached the actor communicates to the source manager to abort the current run.
- the number of events at which evaluate the uncertainty is an estimation of the number of events needed to reach the goal uncertainty. This number is reset at the beginning of each run.

Note: actors only set the fRunTerminationFlag to the source manager in the master thread. However, as this is defined as a global variable and it is therefore seen by the source managers in the other threads,  the abort run signal is sent to all threads at the same time.

Check test066_stop_simulation_criteria_*


opengate/actors/doseactors.py
```diff
@@ -427,18 +427,30 @@ class DoseActor(VoxelDepositActor, g4.GateDoseActor):
...
+            },
+        ),
+        "uncertainty_voxel_edep_threshold": (
             0.7,
             {
-                "doc": "FIXME",
+                "doc": "For the calculation of the mean uncertainty of the edep image, only voxels that are above this fraction of the max edep are considered.",
+            },
+        ),
+        "uncertainty_overshoot_factor_N_events": (
```
nkrah, line 450: Small thing: can you make it a lower case 'n' please?

opengate/actors/doseactors.py
```diff
@@ -427,18 +427,30 @@ class DoseActor(VoxelDepositActor, g4.GateDoseActor):
                 "deactivated": True,
             },
         ),
-        "goal_uncertainty": (
+        "uncertainty_goal": (
             0,
```
nkrah, line 431: I would set the default to None, meaning that the parameter is not set.

opengate/actors/doseactors.py
```diff
@@ -625,6 +639,12 @@ def initialize(self, *args):
         # C++ side has a boolean toWaterFlag and self.score_in == "water" yields True/False
         self.SetToWaterFlag(self.score_in == "water")
 
+        # variables for stop on uncertainty functionality
+        self.SetUncertaintyGoal(self.uncertainty_goal)
```
nkrah, line 643: As a consequence of my other comment, I would add a little logic here: 

```python
if self.uncertainty_goal is None: 
   self.SetUncertaintyGoal(0)
else: 
   self.SetUncertaintyGoal(self.uncertainty_goal)
```
In this way, we can use 0 as indicator on the C++ side that no goal is requested, but use the more user-friendly None on the python side.

opengate/actors/doseactors.py
```diff
@@ -427,18 +427,30 @@ class DoseActor(VoxelDepositActor, g4.GateDoseActor):
...
         ),
-        "thresh_voxel_edep_for_unc_calc": (
+        "uncertainty_first_check_after_n_events": (
+            1e4,
+            {
+                "doc": "Number of events after which uncertainty is evaluated the first time, for each run."
+                "After the first evaluation, the value is updated with an estimation of the N events needed to achieve the target uncertainty.",
+            },
+        ),
+        "uncertainty_voxel_edep_threshold": (
```
nkrah, line 444: I would simply this to *uncertainty_voxel_threshold*

nkrah (commented):
A couple of comments

PR #524, opened 2024-11-05
Snyk clean2

BishopWolf:
Review of Snyk suggestions

### DIFF
--- a/.github/workflows/delocateWindows.py
+++ b/.github/workflows/delocateWindows.py
@@ -48,7 +48,7 @@ def hash_filename(filepath, blocksize=65536):
 
 def find_dll_dependencies(dll_filepath, lib_dir):
     print(dll_filepath)
-    if dll_filepath in global_dll_deps.keys():
+    if dll_filepath in global_dll_deps:
         return global_dll_deps[dll_filepath]
     else:
         dll_deps = {}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -456,6 +456,7 @@ jobs:
           pip install pandas numpy
           python compile_opengate_tests_results.py
     - name: Pushes to another repository
+      continue-on-error: true
       uses: cpina/github-action-push-to-another-repository@main
       env:
           API_TOKEN_GITHUB: ${{ secrets.PUSH_OPENGATE_TESTS_RESULTS }}
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -5,3 +5,6 @@ myst_parser
 sphinx-autoapi
 sphinx_copybutton
 pydata-sphinx-theme
+requests>=2.32.2 # not directly required, pinned by Snyk to avoid a vulnerability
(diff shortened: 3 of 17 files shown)


BishopWolf:
This was way cleaner than previously, however some things remained. Let me know what you disagree on.

nkrah:
Thanks. I’ll have a look.   
  
  
  
  
  
  
  

  
  
>

nkrah:
I had a look. Most changes seem pretty straight forward. Two options: 
1) I create a new branch/PR in the main repo based on your PR and make some minor manual adjustments before merging. We can then do another round of Snyk later. 
2) I push into your branch and we merge based on that. 

Any preference?

BishopWolf:
> Any preference?

Any of the two is fine, if the first is easier for you, then take that route

nkrah:
I'll create a new branch on the main repo tmw.

BishopWolf:
@nkrah I also get these messages

Use binary mode in open (current mode is w+) to avoid encoding-related issues for written file, on Windows or with Python 3. @miscactors.py:177:14 

The issue is that you're opening a file in text mode (`"w+"`) but writing binary data to it. This can lead to encoding-related issues, especially if the data contains non-ASCII characters.

To fix this, you should open the file in binary mode (`"wb+"`) instead. This will ensure that the data is written correctly, without any encoding issues.

Here's the corrected line:

```python
with open(self.get_output_path(which="merged"), "wb+") as f:
```

However, since you're writing JSON data or a string representation of the data, you might want to stick with text mode. In that case, you should ensure that the encoding is correctly set when opening the file. For example:

```python
with open(self.get_output_path(which="merged"), "w+", encoding="utf-8") as f:
```

This will ensure that the file is opened with the correct encoding, and you won't encounter any encoding-related issues.

BishopWolf:
And this one

Explain problem: unguarded~next~call: The call to next should be guarded with a try/except block @delocateWindows.py:134:20

The issue "unguarded `next` call" refers to the fact that the `next` function in Python will raise a `StopIteration` exception if the iterator it is called on is empty. 

In your code, the line `rel_path = next(path for path in pyd_rel_paths if path.endswith(dll))` will raise a `StopIteration` exception if no path in `pyd_rel_paths` ends with `dll`. 

To fix this issue, you should guard the `next` call with a try/except block to handle the `StopIteration` exception. Here's how you can do it:

```python
try:
    rel_path = next(path for path in pyd_rel_paths if path.endswith(dll))
except StopIteration:
    # Handle the case where no path ends with dll
    # You can either raise a custom exception or set a default value for rel_path
    rel_path = None  # or raise ValueError("No path found for dll")
```

Alternatively, you can use the `next` function with a default value, like this:

```python
rel_path = next((path for path in pyd_rel_paths if path.endswith(dll)), None)
```

This will set `rel_path` to `None` if no path ends with `dll`, instead of raising a `StopIteration` exception.

nkrah:
Now in PR #549

PR #538, opened 2024-11-08
remove spurious usages of virtual keyword

alexis-pereda:
This PR removes several usages of the `virtual` keyword.

For those who would want to know why:
[C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override)

By default, if `B` inherits `A` and both provide a member function `f`, calling it through a reference of type `A` or a pointer of type `A*` which actually refers to a `B` (or holds the address of a `B` for the pointer case) will cause the version from class `A` to be used.
Adding `virtual` changes the behaviour to make it so it calls the member function of class `B`.

Making `f` virtual in `A` makes it `virtual` in all inheriting class, so it is not necessary to specify the `virtual` keyword except in class `A`.
However, to indicate in subclasses that a function is `virtual` to the human, before C++11 it was only possible to either add a comment or put the `virtual` keyword anyway.
Since C++11, a new keyword exists: `override`.
It can only be used on member functions that are `virtual`, so the compiler will check that for us (it helps avoiding spelling/type/constness mistakes).

Putting both `virtual` and `override` on the same member function declaration breaks the effect of `override` because even if the function does not exist or is not virtual in the parent class, it will compile as it is marked virtual.

TL; DR: any member function must be marked either `virtual` or `override` (or none), never both.

### DIFF
--- a/core/opengate_core/opengate_lib/GateARFActor.h
+++ b/core/opengate_core/opengate_lib/GateARFActor.h
@@ -23,13 +23,13 @@ public:
   // Constructor
   explicit GateARFActor(py::dict &user_info);
 
-  virtual void InitializeUserInput(py::dict &user_info) override;
+  void InitializeUserInput(py::dict &user_info) override;
 
   // Beginning run callback
-  virtual void BeginOfRunAction(const G4Run * /*run*/) override;
+  void BeginOfRunAction(const G4Run * /*run*/) override;
 
   // End run callback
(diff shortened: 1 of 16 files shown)


dsarrut:
It is tagged as "draft", should we merge or wait?

alexis-pereda:
I was waiting for all the tests to finish to be sure. it can be merged.

PR #548, opened 2024-11-12
New linac design

majacquet:
Update of versa LINAC model


dsarrut:
This PR is 1st priority -> it solves the current issue with the data (see current failed PRs)

PR #550, opened 2024-11-12
Source as a GateObject

dsarrut:
Status:
- DONE: SourceBase, GenericSource, PhaseSpaceSource, VoxelsSource
- DONE: manager and engine
- test004 to test029 are ok

Next steps todo: 
- remove debug/comments
- other sources: beamsources, gansources, phidsources

Warning: this is not the complete refactoring of the source, only make them inherit from GateObject.


nkrah:
@dsarrut Have you seen/considered PR #467?

dsarrut:
> @dsarrut Have you seen/considered PR #467?

nope did not see ... I will merge both

dsarrut:
@nkrah your thoughts: merge or not ?  
Only "big" change : "attached_to" instead of "mother
Also autodoc should work.
(still lot of things todo for refactoring source, but at least the first step is done)

nkrah:
I will have a look
  
  
  
  
  
  
  

  
  
>

nkrah:
I did a quick and shallow review. I think as a first round this should be good and it is really useful because it allows us to use functionality from the GateObject base class. Let's see how the new round of tests comes out.

dsarrut:
we need `add_to_source_manager` as it is reimplemented in PhotonFromIonDecaySource to add all other sub_sources

nkrah:
> we need `add_to_source_manager` as it is reimplemented in PhotonFromIonDecaySource to add all other sub_sources

OK, I see. The method is a bit weird, but I guess we can think about that when we do a second proper round of refactoring ...

PR #555, opened 2024-11-14
Update user_guide_sources.rst

axelrannou:
Add description for direction type (iso, momentum, focused, histogram)


dsarrut:
thanz @axelrannou !

PR #557, opened 2024-11-15
Doc 4 CoinSorter

kochebina:
Documentation for coincidence sorter (simplified version)

### DIFF
--- a/opengate/actors/coincidences.py
+++ b/opengate/actors/coincidences.py
@@ -1,6 +1,71 @@
 import awkward as ak
 from tqdm import tqdm
 from ..exception import fatal
+from .base import ActorBase
+
+class Coincidences(ActorBase):
+    """
+    The Coincidence Sorter searches, into the singles list, for pairs of coincident singles. Whenever two or more singles are found within a coincidence window, these singles are grouped to form a Coincidence event.  All singles open their own coincidence window, and a logical OR is made between all the individual singels to find coincidences.
+    For the moment Coincidence Sorter can  be used only offline. It means that a user should do the simulation in two steps: 1) create system with needed digitizer chain in order to obtain singles within usual Gate simulation;  2) process the file containg the singles in order to obtain coincidecens. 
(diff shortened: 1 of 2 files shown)


opengate/actors/coincidences.py
```diff
@@ -1,6 +1,67 @@
...
+    def initialize(self):
+        ActorBase.initialize(self)
+        if self.policy != "removeMultiples" and self.policy != "keepAll":
+            fatal(
+                f"Error, the policy for the Coincidence Sorter must be removeMultiples or "
+                f"keepAll, while is is '{self.policy}'"
+            )
 
 
 def coincidences_sorter(
```
tbaudier, line 67: Maybe move all following functions into Coincidences class?

opengate/actors/coincidences.py
```diff
@@ -1,6 +1,67 @@
 import awkward as ak
 from tqdm import tqdm
 from ..exception import fatal
+from .base import ActorBase
+
+
+class Coincidences(ActorBase):
```
tbaudier, line 7: Add process_cls(...) at the end of the file
eg: https://github.com/OpenGATE/opengate/blob/master/opengate/actors/arfactors.py#L416

tbaudier, line 7: Add the class to __init__ file: https://github.com/OpenGATE/opengate/blob/master/opengate/actors/__init__.py

tbaudier (changes requested):

opengate/actors/coincidences.py
kochebina: I didn't manage to make it work. Keep it this way for the moment

opengate/actors/coincidences.py
kochebina: Done

kochebina: Done

dsarrut:
sorry, there is still misunderstanding of what a class and an Actor is used for. We will discuss together

PR #569, opened 2024-11-18
Added OptiGAN workflow image

guneet19:
Minor Change: Added OptiGAN workflow image.


nkrah:
Cherry picked this into PR #571

PR #573, opened 2024-11-18
Towards v10

tbaudier:


BishopWolf:
Just 2c

I think for final version 10, Gate should be running on [numpy>=2](https://github.com/OpenGATE/opengate/pull/459) and on [qt6](https://github.com/OpenGATE/opengate/pull/460) in all platforms. This way it will have the same behaviour everywhere. Delaying this will give an immense burden on maintainability.

tbaudier:
Hello,

Use numpy version 2 is not posible for the moment due to pytorch issues.
For qt6, we did not succeed to include the qt librairies in the wheel for the moment. If you have time, maybe you can try to have a look?

For sure, when the 2 PR will be ready, we will merge them and update the Gate version.

PR #574, opened 2024-11-19
UC Davis OptiGAN Bug Fix

guneet19:
Changed OpitGAN code to reflect ` steps_to_store = "first"`.

` steps_to_store = "entering"` leads to incorrect output from OptiGAN. Even filtering the root output gives wrong results. 

@nkrah Is it possible to default to  `steps_to_store = "first"` when using OptiGAN?


guneet19:
@nkrah If you haven't started working on this already. We have a member in our team who can start working on it. He is getting trained on Gate 10 and can be a good starting point for contribution. For now, I will add a doc stating to use steps_to_store = "first" when using OptiGAN. In that case, this code can be merged if there are no conflicts.

nkrah:
@guneet19 : It's on my list, but I have not started yet. Pushing the user guide forward is the priority right now. 

If you can add a line in the doc, as you suggest, that would be good. 

Otherwise, thanks for the offer. Handling the user infos is actually quite deep down in the class factory mechanics used for GateObjects. I don't think that is a good starting point for somebody being trained on GATE 10.

guneet19:
We are working on updated version of optiGAN and found few bugs with kill actor policy. Will update soon.

PR #591, opened 2024-11-21
Compile G4 with a global dynamic TLS model

tbaudier:
It avoid th LD_PRELOAD solution


dsarrut:
... but may be slower (see @andiresch's work). Let's test it before merging

tbaudier:
Here some results. On the ccIn2p3 cluster I run 100 jobs with the version 10.0.1 wheel and 100 jobs with the wheel of that PR (same G4 version)

I run test009_voxels.py

master branch: mean: 1h07min with a standard deviation of 17 minutes
pr branch: mean: 1h012min with a standard deviation of 16 minutes

dsarrut:
Conclusion:

- it seems feasible to avoid the TLS error with this Geant4 flag compilation
- but there is a time penalty of about 10-15%

PR #605, opened 2024-11-26
Stop production actor

andiresch:
This is a translation of Gate9 functionality with test082 examples; 
In the test, the position of the maximum value in the stopping image is compared against the range 50 of the dose actor; The production test, compares against the planned position of a source within the same volume.


andiresch:
General comment/question: There is some redundant code (dose actor, let actor, fluence actor and now also prod-stop-actor) - actually all those image type of scorers. I wonder if we should bring in an intermediate layer of GateVImageActor that handles all the image related stuff - like in Gate9; or we add a few methods (~2-3) to the VActor to reduce levels of inheritences?
Actually, this is not for this PR, but for a future PR;
@nkrah @dsarrut

Issue #590, opened 2024-11-20
Isotopes spectra shall be taken directly from icrp107 database


BishopWolf:
For reference, you can use this code https://www.kaggle.com/code/alexvergaragil/playing-with-radioactive-databases

The icrp107 database is also there [icrp107](https://www.kaggle.com/datasets/alexvergaragil/icrp107/data).

This will add ALL isotopes spectra, not only gamma, but also beta-, beta+, x-ray, etc

BishopWolf:
[icrp107-re.zip](https://github.com/user-attachments/files/17868842/icrp107-re.zip)
here it is the database in json format

PR #608, opened 2024-11-27
Icrp107 database

BishopWolf:
This PR addresses the issue https://github.com/OpenGATE/opengate/issues/590

Basically replace radar database by the ICRP107 database


BishopWolf:
@tbaudier This PR in principle does not change the behaviour of the tests, but the results may change a little bit as a more refined database is used.

BishopWolf:
@tbaudier the precommit bot is doing fancy things. I think I have succeeded this time.

dsarrut:
very good idea, thanks @BishopWolf ! @alexis-pereda can you have a look please to check this is ok for you ?

BishopWolf:
@dsarrut  Everything is working now

BishopWolf:
@alexis-pereda The change in the database makes some test yield a slightly different results, however those should be very close to before. The difference in time it's because the database is bigger, so once it's loaded it should work the same for larger simulations. If some of these presets are not respected, then there is a bug somewhere. 

A solution could be to separate the icrp107.json into components. An icrp107 folder containing 1252 JSON, one for each radionuclide. This way we don't need to read the entire database. Let me know your thoughts. We can off course optimize it.

BishopWolf:
@dsarrut @alexis-pereda I just discovered that icrp107 DO have beta+ spectras. It is just located in the same b-spectra field. therefore you don't even need a separate database for this.

BishopWolf:
Here it is the database separated into components, in binary format to prevent the issues with git transfer
[icrp107.zip](https://github.com/user-attachments/files/17945848/icrp107.zip)

you can read it like
```python
isotopePath = Path("./icrp107") / "Lu-177.json"
with open(isotopePath, "rb") as i:
    isot = json.loads(json.load(i))
    print(isot["name"], isot["emissions"])
```

BishopWolf:
The comparison of icrp107 vs X. Mougeot for F-18 beta+ spectra:

icrp107 mean energy 249.776 keV yield 0.9673 
Mougeot mean energy 250.50 (21) keV

spectra icrp107 (better sampling at lower energies, worse sampling at higher energies)

I think, being Mougeot work more recent it shall be the default and we can rely on icrp107 for the missing isotopes.

|energy (MeV) | weight |
|-------------:|------------:|
|  0.0  |  0.0  |
|  0.0001  |  1.781e-08  |
|  0.00011  |  4.22e-08  |
|  0.00012  |  8.955e-08  |
|  0.00013  |  1.739e-07  |
|  0.00014  |  3.142e-07  |
|  0.00015  |  5.345e-07  |
|  0.00016  |  8.643e-07  |
|  0.00018  |  1.998e-06  |
|  0.0002  |  4.06e-06  |
|  0.00022  |  7.472e-06  |
|  0.00024  |  1.272e-05  |
|  0.00026  |  2.034e-05  |
|  0.00028  |  3.09e-05  |
|  0.0003  |  4.499e-05  |
|  0.00032  |  6.322e-05  |
|  0.00036  |  0.0001145  |
|  0.0004  |  0.0001893  |
|  0.00045  |  0.000323  |
|  0.0005  |  0.000509  |
|  0.00055  |  0.0007551  |
|  0.0006  |  0.001068  |
|  0.00065  |  0.001455  |
|  0.0007  |  0.00192  |
|  0.00075  |  0.002468  |
|  0.0008  |  0.003102  |
|  0.00085  |  0.003824  |
|  0.0009  |  0.004634  |
|  0.001  |  0.006518  |
|  0.0011  |  0.008743  |
|  0.0012  |  0.01129  |
|  0.0013  |  0.01412  |
|  0.0014  |  0.01722  |
|  0.0015  |  0.02055  |
|  0.0016  |  0.02409  |
|  0.0018  |  0.03167  |
|  0.002  |  0.03977  |
|  0.0022  |  0.04827  |
|  0.0024  |  0.05705  |
|  0.0026  |  0.06603  |
|  0.0028  |  0.07515  |
|  0.003  |  0.08436  |
|  0.0032  |  0.09363  |
|  0.0036  |  0.1123  |
|  0.004  |  0.1309  |
|  0.0045  |  0.154  |
|  0.005  |  0.1769  |
|  0.0055  |  0.1995  |
|  0.006  |  0.2218  |
|  0.0065  |  0.2438  |
|  0.007  |  0.2655  |
|  0.0075  |  0.2868  |
|  0.008  |  0.3078  |
|  0.0085  |  0.3285  |
|  0.009  |  0.3488  |
|  0.01  |  0.3887  |
|  0.011  |  0.4274  |
|  0.012  |  0.465  |
|  0.013  |  0.5016  |
|  0.014  |  0.5372  |
|  0.015  |  0.572  |
|  0.016  |  0.6059  |
|  0.018  |  0.6713  |
|  0.02  |  0.7338  |
|  0.022  |  0.7937  |
|  0.024  |  0.8513  |
|  0.026  |  0.9066  |
|  0.028  |  0.9599  |
|  0.03  |  1.011  |
|  0.032  |  1.061  |
|  0.036  |  1.156  |
|  0.04  |  1.245  |
|  0.045  |  1.349  |
|  0.05  |  1.446  |
|  0.055  |  1.537  |
|  0.06  |  1.622  |
|  0.065  |  1.702  |
|  0.07  |  1.777  |
|  0.075  |  1.848  |
|  0.08  |  1.914  |
|  0.085  |  1.976  |
|  0.09  |  2.035  |
|  0.1  |  2.142  |
|  0.11  |  2.236  |
|  0.12  |  2.318  |
|  0.13  |  2.389  |
|  0.14  |  2.449  |
|  0.15  |  2.499  |
|  0.16  |  2.54  |
|  0.18  |  2.594  |
|  0.2  |  2.616  |
|  0.22  |  2.607  |
|  0.24  |  2.57  |
|  0.26  |  2.507  |
|  0.28  |  2.421  |
|  0.3  |  2.314  |
|  0.32  |  2.188  |
|  0.36  |  1.891  |
|  0.4  |  1.549  |
|  0.45  |  1.094  |
|  0.5  |  0.6557  |
|  0.55  |  0.2879  |
|  0.6  |  0.05162  |
|  0.6335  |  0.0  |

alexis-pereda:
I looked at it. It is great
It breaks the `test010_generic_source_energy_spectrum_histogram.py`.
When the source energy type is `"spectrum_histogram"`, the `"energy_bin_edges"` option must be used instead of `"energies"`.
This must contain energy bin edges instead of directly energies, so it will need one more value than the number of weights.

Also, after discussing it with other people, we think we should rename the function `get_icrp107_spectrum` to simply `get_spectrum` instead, but with an extra argument (with default value `"icrp107"`) so that it is more generic and we can possibly add other sources for this kind of data.

BishopWolf:
> so it will need one more value than the number of weights.

@alexis-pereda So if I understand correctly, I must remove the last zero weight if the spectrum type is "b-spectra". Please confirm, this is a very trivial modification.

BishopWolf:
@alexis-pereda Would somethign like this works ? Is it enough?

```python
def set_source_icrp107_energy_spectrum(source, rad):
    if (
        source.particle == "beta-"
        or source.particle == "e-"
        or source.particle == "beta+"
        or source.particle == "e+"
    ):
        rad_spectrum = get_icrp107_spectrum(rad, "b-spectra")
        source.energy.type = "spectrum_histogram"
        source.energy.spectrum_weights = rad_spectrum.weights[:-1]
        source.energy.spectrum_energy_bin_edges= rad_spectrum.energies
    else:
        rad_spectrum = get_icrp107_spectrum(rad, source.particle)
        source.energy.type = "spectrum_discrete"
        source.energy.spectrum_weights = rad_spectrum.weights
        source.energy.spectrum_energies = rad_spectrum.energies
```

alexis-pereda:
If your `rad_spectrum.energies` in the histogram case is correctly providing energy bin edges and not energies, yes.
But I would rather have it named differently as it actually contains different things, even though it is an internal part of the code.

alexis-pereda:
About the trivial modification, I tested it with the `test010_generic_source_energy_spectrum_histogram.py` test, and it still fails.
I am still trying to figure out why because the test expects to get the same values in the root output as it uses for the source.
(so it does not depend on the old data)

BishopWolf:
I obtained 4x speed increase in the tests using individual json files compared to a monolithic database. So I have pushed that. It is also easier to debug and maintain.

BishopWolf:
> I am still trying to figure out why because the test expects to get the same values in the root output as it uses for the source.
> (so it does not depend on the old data)

The test shall replace this function to use the proper initialization

```python
def add_source_energy_spectrum_histogram(sim, phsp, interpolation: str = None):

    source = sim.add_source("GenericSource", "beam")
    source.attached_to = phsp.name
    source.particle = "e-"
    source.n = 5e5 / sim.number_of_threads
    source.position.type = "point"
    source.direction.type = "iso"

    spectrum = set_source_icrp107_energy_spectrum(source, "Lu177") # After defining the particle
    source.energy.spectrum_histogram_interpolation = interpolation

    return source
```

alexis-pereda:
This is not the issue. I did a equivalent change in the test (and to be extra sure, I also tried with this function).
The test runs, it produces a root output, but we get bad results:
test010_plot_energy_spectrum_histogram_None

(you can replicate and find this output graph in `opengate/tests/output/test010/test010_plot_energy_spectrum_histogram_None.png` alongside the produced root file)

BishopWolf:
The problem is this 
image

The old test was made with gammas!!

alexis-pereda:
I already fixed this and used `"b-spectra"` on my local tests.
Do you get correct results with the test on your side?

BishopWolf:
> Do you get correct results with the test on your side?

The old tests are wrong, they use gammas with the energy spectra of electrons. Everything else works fine

alexis-pereda:
Okay, I'll look again with your fix.

(and just to extend a bit on why I am surprised at the tests not passing:
the error in the test in master branch is that it emits gamma particles, but it correctly uses a beta spectrum.
So, while it has to be fixed for users that would use it as a reference, the tests still passes as it only looks at how the energy values are distributed.
Given that, because it only specifically try to match output energies to input energies, I fail to see how it could depend on what exact data is given to the source (as it is the source data that is used in the testing part as "input").
But again, I will see with your update, perhaps I am just missing something obvious here)

BishopWolf:
I see the test 010 passing in all systems now. It remains the test 073 and this is the only thing different:

```
WARNING Compare stats
Runs:         4 4 --> OK? True
Events:       10620854 10146490 : +4.68 %  (tol = 1.00 %) --> OK? False
Tracks:       18329508 17500862 : +4.73 %  (tol = 1.00 %) --> OK? False
Steps:        52593098 50195644 : +4.78 %  (tol = 1.00 %) --> OK? False
PPS:          80573.0 589880.0 : -86.3%    speedup = x0.1 --> OK? True
TPS:          139053.0 1017437.0 : -86.3%    speedup = x0.1 --> OK? True
SPS:          398989.0 2918196.0 : -86.3%    speedup = x0.1 --> OK? True
Track e-      44004 42619 : +3.2% --> OK? False
Track gamma   18285504 17458243 : +4.7% --> OK? False
Tracks      : {'e-': 44004, 'gamma': 18285504} --> OK? True
Tracks (ref): {'e-': '42619', 'gamma': '17458243'} --> OK? True
Tracks vs track_types : 18329508 18329508 --> OK? True
```

In this case it is indeed a database change issue, not much can be done. The difference is not that mch as expected.

BishopWolf:
Test 046 fails on these:

```
Test yield Tc99m: 0.8907654364665489 0.885 False --> OK? False
Test yield Lu177: 0.18033000000000002 0.172168 False --> OK? False
Test yield In111: 1.8474 1.847315 False --> OK? False
Test yield I131: 1.0077017706999998 1.0024600000000004 False --> OK? False
```

It really needs a tolerance, the difference is just the data provenance.

BishopWolf:
test 046 is no longer failing

BishopWolf:
> Also, after discussing it with other people, we think we should rename the function `get_icrp107_spectrum` to simply `get_spectrum` instead, but with an extra argument (with default value `"icrp107"`) so that it is more generic and we can possibly add other sources for this kind of data.

@alexis-pereda I have pushed this change too.

BishopWolf:
@alexis-pereda test 73 discovery Lu177 will always fail since the icrp107 library has 4.74% extra gamma yield for Lu177. This IS a database issue, not a test issue. Therefore, either you rerun the test with the new database or we switch back to radar database.

alexis-pereda:
I think we have to ask @dsarrut on this.

@dsarrut is it possible for you to recreate the reference root file for the test `test073_test5_discovery_lu177_mt.py`
or do you think it is better to keep providing the RADAR source and continue to use it for this test?

BishopWolf:
@alexis-pereda we need a new rebase

BishopWolf:
@dsarrut @alexis-pereda bumping this PR. We need a response to this question:

> @dsarrut is it possible for you to recreate the reference root file for the test `test073_test5_discovery_lu177_mt.py`
> or do you think it is better to keep providing the RADAR source and continue to use it for this test?

dsarrut:
Hi @BishopWolf  @alexis-pereda and happy new year to you ;) 
I think it is better to keep the current test with RADAR data for the moment (for compatibility). 
This is an important PR, let me know if you need further discussion.

BishopWolf:
@alexis-pereda may you investigate why test 72 is failing in windows with py3.11 ?

BishopWolf:
I have created an [issue](https://github.com/radioactivedecay/radioactivedecay/issues/117) in radioactivedacay package to investigate the failing test053

BishopWolf:
@alexis-pereda @dsarrut the package radioactivedecay has been updated with a fix. We must use version 0.6.1

BishopWolf:
It remains this error in MacOS with python 3.12
```
File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/opengate/utility.py", line 9, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

Apparently, the setuptools package is not properly installed in this case. [Read this thread](https://stackoverflow.com/questions/7446187/no-module-named-pkg-resources)

BishopWolf:
Vos données personnelles et celles de l’entreprise sont protégées dans cette conversation

Switching from pkg_resources to importlib.resources in Python can help streamline your code and improve performance. Here’s how you can make the transition:

Using pkg_resources
import pkg_resources

# Example of reading a resource file
resource_content = pkg_resources.resource_string('your_package', 'data/resource.txt')
print(resource_content.decode('utf-8'))

Switching to importlib.resources

For Python 3.7 and later, you can use importlib.resources:

import importlib.resources

# Example of reading a resource file
with importlib.resources.open_text('your_package.data', 'resource.txt') as file:
    resource_content = file.read()
    print(resource_content)

Explanation

Importing the Module:

pkg_resources is replaced by importlib.resources.

Accessing the Resource:

pkg_resources.resource_string is replaced by importlib.resources.open_text.

Reading the Resource:

Use a context manager (with statement) to open and read the resource file.
Benefits of importlib.resources
Performance: importlib.resources is generally faster and more efficient.
Simplicity: The API is simpler and more intuitive.
Standard Library: It’s part of the standard library, so no additional dependencies are required.

By making this switch, your code will be more modern and maintainable. If you have any specific questions or need further assistance, feel free to ask!

Generated with GitHub Copilot

BishopWolf:
@alexis-pereda @tbaudier @dsarrut This PR is ready to merge

tbaudier:
@BishopWolf and @alexis-pereda Are you ok with the last commit?
I created the package https://github.com/OpenGATE/icrp107-database to avoid to have to many files in opengate. And it allows us to import the data without importing opengate.

BishopWolf:
Everything is working fine from my side

tbaudier:
> I rebased on master, it seems ok. Should we squash all commits before the PR in order to reduce the repository size?

Thank you,
Yes please, the squash will remove all occurrences of icrp107 database files in the history.

PR #611, opened 2024-11-28
Update of bias mechanism

majacquet:
Update of GenericBiased tests and add of nonphysicbias method to the PyPhysicList file.


dsarrut:
ok
- [ ] set seed for failing test
- [ ] rename split -> generic

PR #612, opened 2024-11-28
voxelization + attenuation + pytomography

dsarrut:
- some changes for pytomgraphy example
https://github.com/dsarrut/spect_pytomography
- voxelization consolidated (+ IEC)
- attenuation map
- stil WIP


dsarrut:
Still issue with test043_garf on linux only. Unclear why ...

PR #614, opened 2024-11-29
added Gate logo gif animation

andiresch:
the example to simulate the Gate logo using a gaussian parametrization; 
- running multiple runs to make a gif animation;
- by default gets the gaussian parametrization from file
- optional: run a gaussian mixture analysis to parametrize an image; use as input
- added the gate logo to the folder --> maybe this is already in the repo?


dsarrut:
we should put this in the doc also ;)

PR #624, opened 2024-12-05
New image visu

tbaudier:


dsarrut:
This is a trial : with lazy initialisation, the slices seems ok on M1 + QT6. 
The slices were initialized too early. Cleaning to be done

PR #634, opened 2024-12-11
Cancel previous CI jobs for a PR or branch

tbaudier:


.github/workflows/main.yml
```diff
@@ -15,6 +15,10 @@ on:
     - cron: '0 0 * * 0,3'
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.ref }}-ci
```
SimonRit, line 19: Even the master branch?

tbaudier: You are right, thx.
I modify the behavior with the next line: cancel-in-progress: ${{ github.event_name == 'pull_request' }}
I'm not sure how it works with tags

Issue #653, opened 2024-12-27
No coincidences detected when using policy EnergyWeightedCentroidPosition


gertvanhoey:
It seems that there is an issue in GATE 10.0.0 with the policy `EnergyWeightedCentroidPosition` of the `DigitizerAdderActor` if the `DigitizerHitsCollectionActor` is attached to multiple volumes, preventing the detection of coincidences.

In a simulation with two LYSO volumes and a b2b source with n=1, the following two singles are generated when using the policy `EnergyWinnerPosition`. The volume IDs are different and a coincidence is detected.
```
   EventID  PreStepUniqueVolumeID  TrackVolumeName  PostPosition_X  PostPosition_Y  PostPosition_Z  TotalEnergyDeposit  GlobalTime
0        0                    0_0    crystal_rep_0      -19.313315        0.065958        0.108774               0.511    0.059694
1        0                    0_1    crystal_rep_1       23.097943      -15.844351        3.393175               0.511    0.064818
```
However, if the policy `EnergyWeightedCentroidPosition` is used, the following singles are generated. Since they have the same volume ID, no coincidences are detected.
```
   EventID  PreStepUniqueVolumeID  TrackVolumeName  PostPosition_X  PostPosition_Y  PostPosition_Z  TotalEnergyDeposit  GlobalTime
0        0                    0_1    crystal_rep_1      -19.086760        0.078793        0.251413               0.511    0.059694
1        0                    0_1    crystal_rep_1       21.425583       -8.674188        1.444213               0.511    0.064818
```

In the source code `GateDigiAdderInVolume.cpp`, I noticed that `fFinalIndex = i;` is executed for `GateDigitizerAdderActor::AdderPolicy::EnergyWinnerPosition`, but not for  `GateDigitizerAdderActor::AdderPolicy::EnergyWeightedCentroidPosition`. Adding the assignment `fFinalIndex = i;` in the case of `EnergyWeightedCentroidPosition` appeared to solve the issue.

This is the simulation that I was using:
```
import opengate as gate
import uproot
from opengate.actors.coincidences import coincidences_sorter

g_cm3 = gate.g4_units.g_cm3
mm = gate.g4_units.mm
kBq = 1e3 * gate.g4_units.Bq
keV = gate.g4_units.keV
sec = gate.g4_units.second
ns = gate.g4_units.nanosecond
deg = gate.g4_units.degree

def create_materials(sim):
    sim.volume_manager.material_database.add_material_nb_atoms(
[...]

PR #654, opened 2024-12-28
Assign volume index for policy EnergyWeightedCentroidPosition

gertvanhoey:
Fixes OpenGATE/opengate#653


dsarrut:
good catch, thanks a lot !

gertvanhoey:
@dsarrut I want to take the opportunity to say how much I appreciate all the efforts to make GATE Python-based, which makes it so much more accessible and fun to use.

PR #656, opened 2025-01-05
Ensure that volume IDs are unique

### LINKED ISSUE #655: Multiple volumes at the same level in the volume hierarchy result in duplicate volume IDs
I noticed that it was not possible to obtain coincidences between two volumes that were added with two separate `sim.add_volume()` calls. This is caused by using the same copy number `0` in both calls to Geant4's `G4PVPlacement` method. As a result, GATE assigns the same volume ID `0_0` to both volumes and hits appear as though they all happen in the same volume.

gertvanhoey:
Fixes #655

### DIFF
--- a/opengate/engines.py
+++ b/opengate/engines.py
@@ -4,7 +4,7 @@ import sys
 import os
 import weakref
 from box import Box
-from anytree import PreOrderIter
+from anytree import LevelOrderGroupIter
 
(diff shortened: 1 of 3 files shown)


gertvanhoey:
After discussion with @dsarrut it became clear that his was not the best way to solve #655.

Issue #664, opened 2025-01-10
setuptools removed from python 3.12


BishopWolf:
The package setuptools is deprecated and removed from python 3.12. Some distribution may still add it for compatibility, but in general it is gone for good. The user is expected to use importlib instead.

This causes errors in the current CI for MacOS with python 3.12.  Mac users with python 3.12 should be reporting this issue too.

```
File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/opengate/utility.py", line 9, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

Solution, switch to importlib

Switching from pkg_resources to importlib.resources in Python can help streamline your code and improve performance. Here’s how you can make the transition:

# Using pkg_resources

import pkg_resources

## Example of reading a resource file

resource_content = pkg_resources.resource_string('your_package', 'data/resource.txt')
print(resource_content.decode('utf-8'))

# Switching to importlib.resources

For Python 3.7 and later, you can use importlib.resources:

import importlib.resources

## Example of reading a resource file
```
with importlib.resources.open_text('your_package.data', 'resource.txt') as file:
    resource_content = file.read()
    print(resource_content)
```

# Explanation

Importing the Module:

pkg_resources is replaced by importlib.resources.

Accessing the Resource:

pkg_resources.resource_string is replaced by importlib.resources.open_text.

Reading the Resource:

Use a context manager (with statement) to open and read the resource file.
Benefits of importlib.resources
Performance: importlib.resources is generally faster and more efficient.
Simplicity: The API is simpler and more intuitive.
Standard Library: It’s part of the standard library, so no additional dependencies are required.

By making this switch, your code will be more modern and maintainable. If you have any specific questions or need further assistance, feel free to ask!

Generated with GitHub Copilot

BishopWolf:
Update: I tried in Linux with python 3.12, same error
```
Traceback (most recent call last):
  File "/Data/Programas/opengate/bin/opengate_tests", line 5, in <module>
    from opengate.bin.opengate_tests import go
  File "/Data/Programas/opengate/lib/python3.12/site-packages/opengate/__init__.py", line 162, in <module>
    import opengate.sources
  File "/Data/Programas/opengate/lib/python3.12/site-packages/opengate/sources/__init__.py", line 1, in <module>
    from . import generic, phspsources, voxelsources, gansources
[...]

PR #666, opened 2025-01-11
Ensure that volume IDs are unique

### LINKED ISSUE #655: Multiple volumes at the same level in the volume hierarchy result in duplicate volume IDs

gertvanhoey:
Fixes #655

### DIFF
--- a/core/opengate_core/opengate_lib/GateUniqueVolumeID.cpp
+++ b/core/opengate_core/opengate_lib/GateUniqueVolumeID.cpp
@@ -50,7 +50,7 @@ GateUniqueVolumeID::GateUniqueVolumeID(const G4VTouchable *touchable,
     }
   }
   fArrayID = ComputeArrayID(touchable);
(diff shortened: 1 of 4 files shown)


dsarrut:
Excellent Gert, thanks a lot! There is still one failed job, I dont understand why, but it is not related to your changes. I will investigate ...

gertvanhoey:
Happy to be able to contribute! Concerning the failed test, I've seen this in my previous PR too. It was not the same job that failed, but it was also one job out of all the test jobs.

tbaudier:
#664 was merged, I can merge your PR now. Thank you

PR #676, opened 2025-01-29
Reduce contention by allowing simultaneous read access to volume ID map

gertvanhoey:
The method `GateUniqueVolumeIDManager::GetVolumeID` currently requires exclusive access for one thread at a time. But as long as the volume ID map is not modified, it is OK for threads to access it simultaneously. This PR implements a read-write lock to allow for concurrent read access, which can considerably speed up execution in case of multi-threading.


tbaudier:
Hello @gertvanhoey 

I update the master branch to deal with test failing tests. Could you rebase your branch on master please? If you do not know how to do it, a merge could be ok also.
Thank you
Thomas

PR #677, opened 2025-01-31
RE and RBE

andiresch:
This pull request includes:
- relative effectiveness calculations for Alanine including a proton and carbon ion beam test case
- RBE models: mMKM and LEMI low dose approximation, but only a test case for mMKM
- EMCalculator actor to easily retrieve the stopping powers from G4
- Changes in LETActor on c++ side


andiresch:
tests fail due to missing data, but the data are already in the submodule; any idea what's the matter here?

tbaudier:
Thank you very much @andiresch and @MartiDvrk 
Some comments:
 - Is it possible for you to remove that commit from the history: https://github.com/andiresch/opengate/commit/860472d3cbe563f1901d91a86ee86c51da4b1886 It added some binary files and updated subpackages accidentally
 - The subfolder test/data is not updated, this is why the tests are failling
 - Is it possible to rename the 2 files: opengate/tests/src/create_dEdx_table.py and opengate/tests/src/rbe_temp_test.py to test088*_helpers.py or test088*_wip.py for example?

MartiDvrk:
Hi Thomas. I managed to update the submodules and now more tests are passing. However, I still don't have locally some data and I don't understand why. Any idea?

MartiDvrk:
> Hello @MartiDvrk
> 
> The errors for the torch part were resolved with that PR: #679 You can merge the master branch.
> 
> I saw you revert the commit [andiresch@860472d](https://github.com/andiresch/opengate/commit/860472d3cbe563f1901d91a86ee86c51da4b1886) but the problem with a revert is the binary files are still in the history. Is it possible for you to stash all the commits together to avoid that problem? If you do not know how you can do, I can do it :)
> 
> Thank you very much Thomas

Hi Thomas, thank you for the explanation. I merged the fix for torch. I am not very familiar with the stashing. If you have time, it would be great if you could do it. If not, I can take my time to figure out how to do it safely.

dsarrut:
cf #682

PR #686, opened 2025-02-10
remove comments of ` sim.output_dir = paths.output`

jizhang02:
remove comments ((# FIXME (not yet)) ) of ` sim.output_dir = paths.output` in the 3 files of test019. It's already implemented and validated.


dsarrut:
thank you ;) !

PR #693, opened 2025-02-17
Update dataitems.py

majacquet:
Call patch function of itk for image size of [1,1,x] when the uncertainty map is created


dsarrut:
thanks !

PR #698, opened 2025-02-24
Coincidence sorting using pandas functions

gertvanhoey:
Hello @kochebina, this PR contains a modified implementation of the coincidence sorter. 

The modifications include:
* Making use as much as possible of NumPy and pandas functions, to avoid for loops in Python. This results in a big speed increase, which can be x100 or more (the larger the chunk size, the larger the speedup).
* Singles are time-sorted before coincidence detection is done, because I noticed that singles in root files are not always perfectly sorted in time. Especially if the simulation was done with multi-threading, threads typically write singles in groups of 4000, which results in non-monotonous GlobalTime. Because of this, a certain fraction of coincidences may not be detected if the singles are not sorted up front (typically these are random coincidences between singles with different EventID, simulated by different threads).
* While treating chunks of singles, the time window is taken into account and more than one single may be transferred to the next chunk to be used for coincidence detection in the next chunk. I believe that in this way, a small amount of valid additional coincidences may be found.

To test equivalence with the current implementation, I made a script that does the following:
* Run a simulation to generate a singles root file (using multi-threading)
* Read the root file, sort the singles according to GlobalTime and save them to a new root file.
* Process the sorted root file with the current implementation
* Process the unsorted root file with the modified implementation
* Check that both implementations resulted in exactly the same list of coincidences

This script is not part of the PR, but it is available at https://github.com/gertvanhoey/opengate/blob/485db720fb123df108e2af2d948c3b43b7556037/gert/test_coincidence_detection.py

### DIFF
--- a/opengate/actors/coincidences.py
+++ b/opengate/actors/coincidences.py
@@ -1,13 +1,13 @@
-import awkward as ak
+from collections import deque
+from itertools import chain
 
-# import itertools
-# import collections
-import opengate as gate
-from tqdm import tqdm
-from ..exception import fatal
+import awkward as ak
 import numpy as np
-from collections import Counter
-import sys
+import pandas as pd
+
+
+class ChunkSizeTooSmallError(Exception):
+    pass
 
 
 def coincidences_sorter(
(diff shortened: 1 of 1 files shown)


dsarrut:
this is really great, thanks a lot! I think your test script should be integrated as a test. Lets talk about it Friday

PR #707, opened 2025-03-13
Fix RBE LEM1lda bug

MartiDvrk:
Corrections:
- wrong unit in density (cpp side) was leading to results several order of magnitude lower than expected
- beta_mix image is now set to active in the initialization() instead of BeginningOfRun


MartiDvrk:
@dsarrut this is a patch to our previous RBE PR. mMKM is yelding correct results, as it is verified by the test case. However, we could not make a test case for LEM1lda, as we are not allowed to share the tables from our TPS. 
It seems like during the last refactoring of the actor, we made a couple of small mistakes that affect the LEM1lda results and that went undetected. This PR provides a fix.

PR #711, opened 2025-03-20
Patch RBE 

MartiDvrk:
Corrections:

- wrong unit in density (cpp side) was leading to results several order of magnitude lower than expected
- beta_mix image is now set to active in the initialization() instead of BeginningOfRun

Added description of user_input parameters in TP source.


dsarrut:
excellent, thanks ;) !

PR #714, opened 2025-03-21
Remove useless code

tbaudier:
The computation of uncertainty is in the cpp part now


dsarrut:
unsure, to be discussed

PR #718, opened 2025-03-25
DoseActor uncertainty calculation patch

MartiDvrk:
avoid including negative values in the calculation of the mean uncertainty


dsarrut:
unsure, it must never be negative, maybe we should call FlushSquaredValue before ? #719

MartiDvrk:
> unsure, it must never be negative, maybe we should call FlushSquaredValue before ? #719

yes, that could be the problem. I did not understand why the edep square image had value 0 in the voxel, when edep in the same voxel was >0. 
Anyway, even if the check for negative values will be outdated, its current implementation is incorrect.

P.S: I am happy to see how the upcoming Gate meeting is accelerating the solutions of bugs in Gate. It was a good push for us to run many simulations and found out potential issues with our code :)

PR #728, opened 2025-04-02
Set ionization potential

MartiDvrk:
Set the ionization potential of a material via sim.physics_manager.material_ionisation_potential. The user can specify a dictionary with key: material and value: mean excitation energy value to set.
**WARNING:** this doesn't change the value for G4_WATER!

### DIFF
--- a/docs/source/user_guide/user_guide_advanced.rst
+++ b/docs/source/user_guide/user_guide_advanced.rst
@@ -188,15 +188,16 @@ This is a simplification of the true physical process.
 
 Material Ionisation Potential
 ===========================================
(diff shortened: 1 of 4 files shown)


dsarrut:
thanks ! Todo before merge : 
- add documentation (Martina)
- add way to keep the I value even when spawn_new_process is True (David)
@MartiDvrk is the test ready as is ?

PR done during the Hackathon (Athens, 2025) 🎉

MartiDvrk:
> thanks ! Todo before merge :
> 
> * add documentation (Martina)
> * add way to keep the I value even when spawn_new_process is True (David)
>   @MartiDvrk is the test ready as is ?
> 
> PR done during the Hackathon (Athens, 2025) 🎉

I can add the documentation :) The test is ready. We are comparing the range in water simulated with the default I value and with the modified one. Let me know if you have any suggestion to improve it.

PR #729, opened 2025-04-08
H tle dist threshold

majacquet:
Convert the energy cut used to decide whether to apply the TLE (Track Length Estimator) into a corresponding range cut. The type of range cut is determined by the range_type variable.

If range_type = "max", the range cut is estimated as the CSDA (Continuous Slowing Down Approximation) range of an electron produced by a photon transferring all its energy.

If range_type = "average", the range cut is estimated as the CSDA range of an electron with the average energy expected from a photon undergoing a process.

The GateUserTrackingInformation was also improved to provide an actor dependent way to provide information (for the moment only boolean) on the track.


majacquet:
I redo my speed comparison calculation, and for the worst possible case : each voxel = diff material and  source within the volume, the max range TLE is 9 % slower than the ekin threshold TLE, whereas the average range TLE is 16% slower than the ekin.

PR #739, opened 2025-04-23
read_phantom_mhd_image

Varvara1990:
This is an example of how to read an mhd phantom image and locate it in the middle of the world. 
It is needed two additional files (UncompressedBreast22.mhd,  UncompressedBreast22.raw) in order to be able to work which I am not able to upload since their type is not supported.


dsarrut:
not a PR ;)

PR #745, opened 2025-04-30
Add test 092 for optic material issue

tbaudier:
The problem comes from the material names created by opengate in the volume does not have the same name than the material in the db


tbaudier:
Done here https://github.com/OpenGATE/opengate/pull/780

PR #755, opened 2025-05-13
Generic filters

alexis-pereda:
A new system to configure filters for actors.

The main idea is to be able to write for example:
```py
actors.filter = "kinetic_energy < 20 * MeV or particle_name == 'gamma' and track_id != 0"
```

Current proof of concept works, run time overhead yet to assess to compare to another implementation (partial) using C++ JIT compilation (preliminary tests show a factor of around 2 for short simulations on m'y computer).
To discuss : optional dependency on (py)ROOT (for C++ JIT) to reduce simulation time when using generic filters.

Demo in test023 generic (not an actual test yet, must be executed with options, e.g. -san1e2).


opengate/tests/src/output/dose_dose.raw
tbaudier, line 1: binary file to move to data

opengate/tests/src/output/dose_dose_uncertainty.raw
tbaudier, line 1: binary file to move to data

opengate/tests/src/output/dose_edep.raw
tbaudier, line 1: binary file to move to data

alexis-pereda: (for all) yes, no problem. These files probably won't exist in a few commits anyway, but I will follow this for the files that will be produced by the actual test.

Issue #719, opened 2025-03-25
bug uncertainty on the fly


dsarrut:
https://github.com/OpenGATE/opengate/blob/f1ae1c9882374fb50a16eec271e191abe09644d6/core/opengate_core/opengate_lib/GateDoseActor.cpp#L264

potential bug : need to Flush before !

MartiDvrk:
I was thinking that we could flush the thread local 1D images into the 3D edep_squared image, as it is done at the end of run. Additionally, we would need to reset the local thread images to zero, to avoid adding the values twice. What do you think?

PR #757, opened 2025-05-14
PR for issue #719

MartiDvrk:
- Call FlushSquaredValue() explicitly at the beginning of EndOfEventAction(). In fact, there is an implicit flushing implemented in ScoreSquaredValue() triggered in SteppingAction that flushes deposit in a voxel whenever a deposit by a new event is made. Consequently, the deposit made by the last event in any voxel remains unflushed and we must trigger the flushing explicitly. 
- At the same time, once the value is flushed, we must make sure that the voxel in the thread local image is reset to zero. So at the end of FlushSquaredValue(), we must iterate over the image and reset all pixels to zero.

### DIFF
--- a/core/opengate_core/opengate_lib/GateDoseActor.cpp
+++ b/core/opengate_core/opengate_lib/GateDoseActor.cpp
@@ -246,7 +246,19 @@ void GateDoseActor::SteppingAction(G4Step *step) {
 
 void GateDoseActor::EndOfEventAction(const G4Event *event) {
 
-  // flush thread local data into global image (postponed for now)
+  // flush thread local data into global image
+  // reset local data to zero
+  int N_voxels = size_edep[0] * size_edep[1] * size_edep[2];
+  if (fEdepSquaredFlag) {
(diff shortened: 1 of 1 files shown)


dsarrut:
not sure : the flush is implicit each event and should NOT being explicit for all the pixels at each event, otherwise it will be too long. The flush should be only 1) at the end of run or when you compute the uncertainty, within the "if NbOfEvent >= NbEventsNextCheck" part. 
Indeed, the 1D data must be set to zero within FlushSquaredValue () event after each Run

dsarrut:
seems good ! An additional  change please (the last ;) !). Can you set to zero all the values of squared_worker_flatimg in FlushSquaredValue, after the end of the loop ? I think when there are several runs or when the uncertainty is computed, it must be set to zero to avoid "flushing" two times the values in a subsequent event. Thanks ! 
PS: dont bother about the failing test "test081_simulation_optigan_with_random_seed" it is a knowing issue, under investigation.

PR #759, opened 2025-05-15
Stopping Actor compatible with filters

MartiDvrk:
The GateVActor PostUserTrackingAction was overwritten by the GateProductionAndStoppingActor. We need to call the mother class function to be able to use the actor with filters.
@yihan-jia


dsarrut:
excellent, thanks @MartiDvrk and @yihan-jia ;) !

PR #772, opened 2025-06-09
Add possibility to output coincidences to file

gertvanhoey:


gertvanhoey:
@dsarrut @kochebina @tbaudier I am working on a modification of coincidence sorter that allows to save coincidences to a file instead of keeping them in memory until the end.

What do you think about the following?
* It seems that uproot does not allow to update a root file with additional data. So with root as the output file format, it would still be needed to keep all coincidences in memory until the end. So I chose HDF5 as the output format instead. Is this a good choice for GATE?
* I'm assuming that we want backward compatibility (existing code that uses `coincidence_sorter()` does not break), so the function still returns coincidences in memory (Python dictionary or pandas DataFrame) by default, but saving to a file can be selected with two additional arguments (boolean + file path). This function interface is maybe not so clean. An alternative is to provide an additional function specifically for coincidence sorting with file output.

dsarrut:
thank a lot Gert ! Yes HDF5 is a good choice (we plan to support more HDF5 in the future). However, for compatibility, do you think it is possible to convert a HDF5 to a root file once the program is completed ?

> @dsarrut @kochebina @tbaudier I am working on a modification of coincidence sorter that allows to save coincidences to a file instead of keeping them in memory until the end.
> 
> What do you think about the following?
> 
> * It seems that uproot does not allow to update a root file with additional data. So with root as the output file format, it would still be needed to keep all coincidences in memory until the end. So I chose HDF5 as the output format instead. Is this a good choice for GATE?
> * I'm assuming that we want backward compatibility (existing code that uses `coincidence_sorter()` does not break), so the function still returns coincidences in memory (Python dictionary or pandas DataFrame) by default, but saving to a file can be selected with two additional arguments (boolean + file path). This function interface is maybe not so clean. An alternative is to provide an additional function specifically for coincidence sorting with file output.

kochebina:
Dear @gertvanhoey, thank you a lot!
I'd say that eventually we must have root output as people doing PET usually are very used to it. I agree with @dsarrut that maybe we can have (automatic) converter at the end of execution? Maybe we can un option which format user would like HDF5, ROOT (converted from HDF5, increase in exe time but that's life), binary, CASTor maybe ? 
@dsarrut what do you think?

gertvanhoey:
@dsarrut @kochebina @tbaudier 
I updated the Coincidence Sorter documentation as part of this PR. I was wondering: do we want to keep the note that the coincidence sorter is a work in progress because it can only be used offline?

gertvanhoey:
@tbaudier The unit test that I added fails with Python 3.9 (on Windows) because saving to a HDF5 file in append mode requires pytables>=3.10 if numpy>=2.0 is used (https://github.com/PyTables/PyTables/issues/1172), but that version of pytables is only supported from Python 3.10 onwards (https://github.com/PyTables/PyTables/blob/v3.10.0/pyproject.toml). Why the test still succeeds on Linux with Python 3.9 is not clear to me, though.
What do you think would be best to deal with this? I could remove support for HDF5 output again from the coincidence sorter, but maybe you see other possibilities?

tbaudier:
Hello,
So bad, maybe one idea is: just before to save the hdf5 file, check the version of python. If 3.9, raise an exception with a message for the user saying to use python 3.10 at least for hdf5.
By the way, python 3.9 won't be supported anymore at the end of the year.

gertvanhoey:
@tbaudier Thank you for the suggestion. I made this modification and I excluded part of the unit test in Python 3.9.

tbaudier:
It's perfect, thank you. I merge it

PR #774, opened 2025-06-11
Add support of OFF files for tessellated volumes

alexis-pereda:
This PR adds [Object File Format (OFF)](https://en.wikipedia.org/wiki/OFF_(file_format)) to the supported file formats for tessellated volumes (previously only STL).

It also adds an option to these volume to disable the centering on the center of gravity (the current behaviour is kept as default).

The associated test (`test067_tesselated_volume.py`) is split into two:
- `test067_tesselated_stl_volume.py` (the current one);
- `test067_tesselated_off_volume.py`, mainly a copy of the STL one but with an OFF file.

### DIFF
--- a/docs/source/developer_guide/developer_guide_notes.rst
+++ b/docs/source/developer_guide/developer_guide_notes.rst
@@ -102,7 +102,7 @@ So, after a long explanation, the key points are:
    release_g4_references() method is called by the class’s close()
    method.
 
-An example may be the implementation of the Tesselated Volume (STL).
+An example may be the implementation of the Tesselated Volume (STL or OFF).
 
 Pybindings in pyG4TriangularFacet.cpp:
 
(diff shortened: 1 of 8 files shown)


alexis-pereda:
Good point, thank you both!
I replaced my code and the previous code for STL with calls to meshio.
I updated the documentation accordingly.

@tbaudier I think we should add meshio to the dependencies of opengate, should I let you do it?

Also, I noticed that "Tesselated" is not spelled correctly (it should have two 'l'). This is not the point of this PR, but I wanted to have other people's viewpoint on whether we should correct it (and invalidate existing code using the current name) or not.

BishopWolf:
All tests are failing because meshio is not being installed

alexis-pereda:
I added meshio in the requirements.
I only kept the original test associated with tessellated volumes and it generates a STL file on the fly using `numpy-stl`.
I can rewrite it to remove this dependency, if you think it is relevant.

alexis-pereda:
I added meshio in the requirements.
I only kept the original test associated with tessellated volumes and it generates a STL file on the fly using `numpy-stl`.
I can rewrite it to remove this dependency, if you think it is relevant.

PR #778, opened 2025-06-17
Add air in GateMaterials.db in contrib

tbaudier:
Air is necessary for some other application eg: pCT


SimonRit:
Shouldn't one use G4_AIR instead?

PR #835, opened 2025-11-05
Fix user limits (step limiter)

nkrah:
This PR fixes a performance issue observed in carbon ion therapy simulations where the dose actor could become unexpectedly slow because the step limiter was not applied to all relevant particles.

Previously, GATE treated the particles eligible for UserLimits similarly to production cuts and effectively restricted step-limiter selection to protons, electrons, positrons, and gammas. Geant4 can apply UserLimits to any particle with a process manager, and GATE v9 allowed this broader behavior. As a result, GATE v9 and GATE v10 applied step limiters differently, which could change tracking behavior and degrade performance in the affected simulations.

The main change updates the handling of `sim.physics_manager.user_limits_particles` so UserLimits can be applied to arbitrary Geant4 particles, including ions such as `GenericIon`. The special selectors now have explicit semantics:

- `all` applies UserLimits to all particles known to the Geant4 particle table.
- `all_charged` follows Geant4's step-limiter behavior for charged particles.
- Historical aliases such as `electron` and `positron` are translated to their Geant4 names.

The PR also updates tests that previously relied on the old `all` behavior. Where a test should keep the historical four-particle selection, it now explicitly requests `proton`, `gamma`, `electron`, and `positron`. The user and developer documentation were updated to describe the new API and the internal step-limiter implementation.

The temporary debug commits used during the investigation have been removed before merging.

### DIFF
--- a/docs/source/developer_guide/developer_guide_physics.rst
+++ b/docs/source/developer_guide/developer_guide_physics.rst
@@ -1,0 +1,161 @@
+Physics
+=======
+
+This chapter documents implementation details of GATE's physics handling for
+developers. It focuses on the Python-side orchestration around Geant4 physics
+objects, i.e. the code that turns user-facing settings into Geant4 regions,
+physics constructors, processes, and cuts.
+
+Step Limiter
+------------
+
+The step limiter is implemented through Geant4 user limits. On the GATE side,
+the relevant code is split across three places:
+
+* ``opengate.managers.PhysicsManager`` exposes the user-facing API.
+* ``opengate.physics.Region`` stores per-region user limits and builds the
+  corresponding Geant4 region objects.
+* ``opengate.physics.UserLimitsPhysics`` is a Geant4 physics constructor that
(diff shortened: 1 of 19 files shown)


core/opengate_core/opengate_lib/GateSimulationStatisticsActor.cpp
```diff
@@ -103,6 +108,17 @@ void GateSimulationStatisticsActor::PreUserTrackingAction(
   if (fTrackTypesFlag) {
     auto p = track->GetParticleDefinition()->GetParticleName();
     data.fTrackTypes[p]++;
+
+    // For DEBUGGING only - remove
```
BishopWolf: https://stackoverflow.com/a/561338/1103558
You can check if you're in a debug environment; this will be ignored on release.

dsarrut:
replace test where step_limiter = "all" -> by "e- p etc..."

PR #845, opened 2025-11-17
Update Github CI

tbaudier:
Remove python 3.9 EOL
Add python 3.13
Remove macos CI for PR to limit build time
Add Windows ARM for tag version


BishopWolf:
@tbaudier Add Python 3.14, as it is the default now for a lot of systems

PR #850, opened 2025-11-25
Corrected the TAC interpolation function GateGenericSource::UpdateActivityWithTAC

Marc-AntoineLeclerc:
GateGenericSource::UpdateActivityWithTAC in opengate/core/opengate_core/opengate_lib/GateGenericSource.cpp does not interpolate the TAC_activities correctly. There are two issues:

1. The index i is not calculated properly. It is one index too high.
2. The interpolation equation is flipped. w1 and w2 should be swapped.

I corrected both issues.


dsarrut:
excellent, thank you very much!

PR #852, opened 2025-11-28
Coresi ccmod

dsarrut:
- macaco1 test1 with singles : DONE
- macaco1 test1 with coincidences : almost
- starting coresi interface: still no test yet


dsarrut:
See #996

PR #878, opened 2026-01-08
Add castor tests during CI

tbaudier:
Add Docker file to create the docker image. Has to be manually created due to registration for gitlab access


tbaudier:
python test096_pet_castor.py -> test the pet output
python test096_pet_castor_ref.py -> same than previous but with more particles for the next part
python test096_pet_castor_coinc.py -> create coincidences

PR #895, opened 2026-02-04
Add proton CT runnable example

acoussat:
Proton CT runnable example. The goal is to use it in connection with [PCT](https://github.com/RTKConsortium/PCT) in order to have a complete proton CT pipeline (from MC simulation to tomographic reconstruction) easily runnable.

The example is [taken from PCT](https://github.com/RTKConsortium/PCT/blob/6afe8ee0a0c25fd3e892761a237f120fd19198e4/gate/protonct.py), from which it will probably be removed. This example was already somewhat available in GATE in the form of a [GATE 9 to GATE 10 migration guide](https://github.com/OpenGATE/opengate/blob/b2d485fd51cce5af3de600d05eb153a409c143a7/docs/source/user_guide/user_guide_how_to_convert_example_1.rst).


SimonRit (approved):

nkrah:
I agree that AddTrackerUncertainty.py should be part of PCT. It is conceptually closer to the reconstruction than to the simulation. 

@acoussat : Otherwise, what is the status of this and the sibling PR?

acoussat:
I have just finished the [sibling PR](https://github.com/RTKConsortium/PCT/pull/75), so if everyone agrees this one is ready for review/merge!

pyproject.toml
dsarrut: unsure protonct is a script here. Look like an example. We put as script tools (usually with options) that perform a Gate related task (compute dose rate, extract gammas etc).

dsarrut:
This example does not seems to validate anything, no ? Should be a "test" that compare the output to a reference (or at least, a previous validated output)

acoussat:
I added a basic test that compares the output of the small simulation to a reference.

PR #905, opened 2026-02-12
Dose Actor speed up

MartiDvrk:
This PR implements:
1) scoring of dose to any material (not only dose to water) -> test041_dose_actor_dose_to_other_material_mt.py
2) possibility to speed up the dose to dose to material conversion by assuming the stopping power ratio (SPR) constant with energy. ->test041_dose_actor_SPR_cache_mt.py

In 2), the SPR is calculated only once per particle-voxel material pair, and stored in a cache. At every step, the dose actor queries the cache for the SPR value, given the current interacting particle and voxel material, in a "FindOrCalculate" fashion.

This mechanism can be activated by the user, by setting the `assume_constant_SPR_per_material` flag to True in the dose actor. The user can also specify the energy to be used for the SPR calculations, with `constant_energy_SPR`.

This method is based on the fact that, for protons, the SPR (S in water / S in material) is more or less independent on energy, for E > 10 MeV, and depends only on the voxel's material. See "Dose to water versus dose to medium in proton beam therapy", Paganetti, 2009.
For SPR in material other than water, and particles different from protons, the user is invited to verify the applicability of the method.


BishopWolf:
This is for C-12 ions, but it holds for protons, it is not really that constant
image

MartiDvrk:
> Isn't it better to implement a dictionary based on this publication? https://www-pub.iaea.org/MTCD/Publications/PDF/p15048-DOC-010-398-Rev1_web.pdf Specifically table 22 gives you the stopping power ratio water/air for electrons
> 
> EDIT: I incorrectly assumed you used a fixed constant value. I can see you calculate it first. But this shall use the energy already declared by the user and not an extra energy value. The "constant_energy_SPR" shall be extracted from the user input

Hi @BishopWolf  :) The constant_energy_SPR value, provided by the user, is already used to set up the energy value used for the calculation. In the python DoseActor, in the "initialize" function, we call `self.SetConstEnergyForSPR(self.constant_energy_SPR)`. This function is setting the user input to the DoseActor in cpp, where it is used for the SPR calculation.

PR #907, opened 2026-02-15
Define intrinsic activity (e.g., Lu176 background) confined specifically to the crystal volumes.

AmirZamanpourKUL:
Confine the activity to last crystal layer without repetition.


nkrah:
Sorry for not considering this PR earlier. It is a very useful illustration of how to confine a source to a repeated volume. Two things: 

1) Is this test linked to the documentation about source confinement? -> would be useful. 
2) The test should actually verify that the source confinement works as expected. For that, I would simplify the test: one mother volume, 2 daughter volumes. PhaseSpaceActor, attached to mother, scores first steps. Based on phase space actor output, verify that particles are produced in both sub volumes and confined to them.

PR #913, opened 2026-02-24
Dev fluence actor

Giorgio5i6:
Implemented the first version of the fluence actor as an extension of the original one.
The user can select which actor to use: myFluenceActor.fluence_scoring_mode = "sum_tracks" #(or "fluence").
Part of this scorer was adapted from the dose actor.

TO DO: 
1. systematic comparison with independent Monte Carlo
2. Improve consistency with other scorers
3. Make tests for this scorer


tbaudier:
Hello @Giorgio5i6 
The Fluence actor has been merged recently with that PR: https://github.com/OpenGATE/opengate/pull/989
Do you have everything with this new part? Or do you think some functionality are missing?

PR #917, opened 2026-02-27
Coincidences fixes

gertvanhoey:
@dsarrut @kochebina
Non-monotonicities in the GlobalTime (e.g. because of divergence of GlobalTime in multiple simulation threads) was not properly handled in coincidences.py. This caused the problems that Marcin was running into lately. This has been fixed in this PR. I also fixed code related to writing root output, which was no longer supported since the recent uproot 5.6.7.


tbaudier:
Hello @gertvanhoey 
Could you rebase your PR please. We merge G4 11.4.0 in the master branch. Thank you

Issue #928, opened 2026-03-05
Simplify OpenGate Core wheel generation


BishopWolf:
We currently have three ways to generate wheels, one for each operative system.
However, cibuildwheel offers a unified API to do this (https://deepwiki.com/pypa/cibuildwheel/4.1-github-actions-integration).
The API also handles delocation by itself.

I propose to simplify the opengate_core wheel generation, which will improve the maintainability by reducing the complexity. The change must keep the current flow.

EDIT: In case it is not clear:

Current status: You build geant4 and ITK for every python version and for each system.

With the proposed change: You build geant4 and ITK ONCE per system, then you create opengate_core wheels for each python version. You still need to compile for each system, but the total compilation time will decrease like 70%

BishopWolf:
@tbaudier question
I understand bundling qt6 in windows and mac, OK.

I don't understand the need to bundle qt6 in Linux, we can ask the user to install qt6 if he has not installed it. We can always use the system's qt

BishopWolf:
@tbaudier @dsarrut @nkrah 

PR #935 is complete
All wheels are being produced
All tests are passing

Original goals:

- [x] Move all wheel generation to use cibuildwheel
- [x] Simplify all the scripts and possibly avoid them
- [x] Keep current functionality as is (bundled qt6, geant4 and ITK)
- [x] Remove manual delocation completely, let cibuildwheel do all the work

Missed goals, moved for a new PR:

- [ ] Add qt6 to windows
- [ ] Preinstall qt6 in windows and macos in a cache

PR #935, opened 2026-03-09
Use cibuildwheel to automatically build all the wheels, avoid manual builds

BishopWolf:
This PR is intended to open a discussion to address the issue https://github.com/OpenGATE/opengate/issues/928

This changes current wheel generation. **CONFIRM AND APROVE BEFORE MERGE!!!**

- [x] Windows is producing all the wheels now, fully auto using delvewheel
- [x] Linux is producing all the wheels now, fully auto
- [x] Macos is producing all the wheels now, fully auto
- [x] All Tests passing
-  Add Qt to Windows. **Will be made in another PR**
-  cache brew for macos, this will reduce build time. **Will be made in another PR**

# Current status: 

* Supported ARCHS: 
    * Windows x86_64 (NO ARM !)
    * Linux x86_64 ; ARM
    * MacOS intel ; ARM64
* Supported Python: 3.10 up to 3.14; except macos intel which has no python 3.14 support
* Currently we build one wheel per python version and per system.
* The selection of python versions is manual. But we start with a pinned version, then we let CIBW build the correct wheel for the desired version.
* Total build time averages 45 minutes before tests. Each build takes 16 minutes preparation and around 15 minutes build, but not all start at the same time. 

# Possible improvements:
* We could let cibuildwheel select the python version automagically. This is scalable in time as cibuildwheel knows which python versions correspond to each system.
* We could use a single system to build the tagged wheels, this will simplify a lot the action scripts. Even for novis builds you can just use python 3.12 to build the wheel for 3.9 if needed.
* Total build time will average 50 minutes dropping unneeded python versions. 16 minutes preparation ONCE and 10 minutes per build for each python version. In the case of macos, the delocation also takes 10 more minutes, its almost instantaneous in linux and windows.


BishopWolf:
I have rebased the last working commit to current master. I can add the intended modifications in a different PR.

Current issues:
When adding QT in windows : the virtual machine can't find the conda location. https://github.com/BishopWolf/opengate/tree/windows-qt6

When using homebrew cache in mac : cibuildwheel can't find system python

BishopWolf:
@tbaudier @dsarrut @nkrah Let me know when are you available to discuss this PR

tbaudier:
Hello @BishopWolf can you rebase the PR please, specially with the new version tag, your commit [c616243](https://github.com/OpenGATE/opengate/pull/935/commits/c616243797f877c387817134ec0bef7faee9fcb0) should be removed. Then I will test the visualization part. Thank you

core/VERSION
```diff
@@ -0,0 +1 @@
+10.1.0
```
tbaudier: File to remove, the VERSION file is already present in opengate folder. It's better to avoid duplication

BishopWolf: cibuildwheel needs a self contained folder for the build. This file is not present in the container.

A solution would be to copy it before the build stage

BishopWolf:
@tbaudier I have also tested in a macos-15-intel I have, and it builds fine with passing tests on python 3.12

tbaudier:
Thank you, I will do some test because I have a macos intel to check the visualisation. I let you know

BishopWolf:
@tbaudier 
for the ci in macos intel and windows arm we can set miniconda with python 3.12, i don't think there's a higher version supported

EDIT: Confirmed for these OS: Miniconda 26.3.2-2
Apr 28, 2026 https://www.anaconda.com/docs/getting-started/miniconda/release/26.x
Available Python versions: 3.13.13, 3.12.13, 3.11.15, 3.10.20

So no python 3.14

BishopWolf:
In my macos intell brew can install qt, even in the ci it could. But somehow delocate can't find it. This is weird, i'll revert to conda qt

BishopWolf:
@tbaudier windows arm is not suppported by conda https://repo.anaconda.com/miniconda/
either we find a replacement or we drop the support for it

BishopWolf:
@tbaudier everything is ready to merge. You can do your tests

BishopWolf:
@tbaudier I really think the issue of brew qt is that geant4 was compiled with conda qt and it doesn't like the brew one.

Either we recompile geant4 with brew qt as in macos arm64 or we keep conda for qt in intel which adds extra maintenance burden

To keep conda just delete the last two commits

BishopWolf:
@tbaudier since I will change qt installation in another PR, it makes no sense to have it here. I have reverted the last two commits.

BishopWolf:
@tbaudier macos intel builds fine locally, i don't understand well the errors in the build, it complains about iconv and cups libraries

tbaudier:
test visualization:

| OS | misc/test004_simple_visu_qt.py | misc/test004_simple_visu_qt_volume_style.py | geometry/test009_voxels_visu.py | geometry/test009_voxels_plan_homothety_visu.py | 
|----------|----------|----------|----------|----------|
| macos intel | ok | ok | ok | ok |
| macos MX | segfault | | | |
| Linux | segfault (ubuntu) but also with pypi version | | | |
| window | | | | |

PR #975, opened 2026-03-28
Fix PhaseSpaceActor scoring in nested attached volumes

nkrah:
Summary:
- add a GateVActor helper to detect whether a step occurs in the top attached volume rather than a daughter
- use that helper in GatePhaseSpaceActor to ignore subtree-triggered internal surface crossings

Motivation:
PhaseSpaceActor is attached to a single volume in GATE, but the underlying sensitive detector registration is recursive over the attached volume hierarchy. This can lead to surprising extra phase-space entries when daughter volumes are crossed.

Changes:
- add GateVActor::IsStepInTopAttachedVolume(const G4Step*)
- early-return in GatePhaseSpaceActor::SteppingAction() unless the step is in the top attached volume

Notes:
- this keeps the framework helper isolated from the actor behavior change in two separate commits
- it does not change recursive SD registration globally; it only tightens PhaseSpaceActor behavior

### DIFF
--- a/core/opengate_core/opengate_lib/GatePhaseSpaceActor.cpp
+++ b/core/opengate_core/opengate_lib/GatePhaseSpaceActor.cpp
@@ -132,6 +132,15 @@ void GatePhaseSpaceActor::SteppingAction(G4Step *step) {
 
   auto &l = fThreadLocalData.Get();
 
+  // Check if this SteppingAction was triggered via the sensitive detector 
+  // attached to the top volume, i.e. the "attached_to" volume
+  // if it was triggered by a daughter volume, this is not truely a step
+  // at the outer surface, but rather an irrelevant (for the purpose of this actor) 
+  // internal surface crossing
+  if (!IsStepInTopAttachedVolume(step))
+    return;
+
+
   // Particle enters the volume if the pre step is at the volume boundary
   const bool entering =
       step->GetPreStepPoint()->GetStepStatus() == fGeomBoundary;
--- a/core/opengate_core/opengate_lib/GatePhaseSpaceActor.cpp
+++ b/core/opengate_core/opengate_lib/GatePhaseSpaceActor.cpp
@@ -132,14 +132,13 @@ void GatePhaseSpaceActor::SteppingAction(G4Step *step) {
 
   auto &l = fThreadLocalData.Get();
 
-  // Check if this SteppingAction was triggered via the sensitive detector 
+  // Check if this SteppingAction was triggered via the sensitive detector
   // attached to the top volume, i.e. the "attached_to" volume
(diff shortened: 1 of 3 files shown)


tbaudier:
Hello @nkrah 
I think your PR modify the behavior of the test: geometry/test058_iec_six_spheres.py

core/opengate_core/opengate_lib/GateVActor.cpp
```diff
@@ -175,6 +175,28 @@ bool GateVActor::IsStepEnteringVolume(
...
 }
 
+bool GateVActor::IsStepInTopAttachedVolume(const G4Step *step) const {
+  const auto *touchable = step->GetPreStepPoint()->GetTouchable();
+  if (touchable == nullptr)
+    return false;
+
+  const auto history_depth = touchable->GetHistoryDepth();
+  int attached_volume_depth = -1;
+  for (int i = 0; i <= history_depth; i++) {
```
tbaudier, line 185: Why did you do a for loop instead of looking for the name of  touchable->GetVolume(0)?

core/opengate_core/opengate_lib/GatePhaseSpaceActor.cpp
```diff
@@ -132,6 +132,14 @@ void GatePhaseSpaceActor::SteppingAction(G4Step *step) {
...
   auto &l = fThreadLocalData.Get();
 
+  // Check if this SteppingAction was triggered via the sensitive detector
+  // attached to the top volume, i.e. the "attached_to" volume
+  // if it was triggered by a daughter volume, this is not truely a step
+  // at the outer surface, but rather an irrelevant (for the purpose of this
+  // actor) internal surface crossing
+  if (!IsStepInTopAttachedVolume(step))
+    return;
+
```
tbaudier, line 142: How can we manage the others policy of the storing: ‘exiting’, ‘first’ or ‘all’

core/opengate_core/opengate_lib/GatePhaseSpaceActor.cpp
```diff
@@ -132,6 +132,14 @@ void GatePhaseSpaceActor::SteppingAction(G4Step *step) {
 
   auto &l = fThreadLocalData.Get();
 
+  // Check if this SteppingAction was triggered via the sensitive detector
+  // attached to the top volume, i.e. the "attached_to" volume
+  // if it was triggered by a daughter volume, this is not truely a step
```
tbaudier, line 137: We need to pay attention to daughter volume with same boundary than mother volume when the particle enters.
Or if the particle is created inside a daughter volume like in test058 with iec spheres, the mother volume is the world

PR #977, opened 2026-03-30
Add running test for linac to be used on VIP

tbaudier:


tbaudier:
For the moment the test can be run with:
```
docker run --rm tbaudier/opengate_vip:0.1.1 /opt/python/cp312-cp312/lib/python3.12/site-packages/opengate/tests/src/advanced_tests/linacs/run_linac_test.sh
```
without parameters and lasts around 3h.
The docker image is available here: https://hub.docker.com/repository/docker/tbaudier/opengate_vip/general

PR #1001, opened 2026-04-17
Cluster dose

nkrah:
New actor to score ionization cluster dose. 

Work in progress with Krakow team.


nkrah:
will work on this in a fork first

PR #1006, opened 2026-04-23
Add DepositedChargeActor

srmarcballestero:
## Rationale

This PR adds an actor to score the net electric charge deposited in a volume during a simulation. I have named it `DepositedChargeActor`.

## How does it work?

When a track is born inside the volume, it contributes -q to a running counter, when it dies inside the volume, it contributes +q. The value of this counter at the end of the simulation is  the net deposited charge. Each contribution is weighted by the track weight, so the totals remain correct under biasing or weighted sources.

Two quantities are recorded:

- `deposited_nominal_charge`: scores the charge of the particles according to `G4ParticleDefinition::GetPDGCharge`.
- `deposited_dynamic_charge`: scores the charge of the particles according to `G4DynamicParticle::GetCharge`, which accounts for the ionisation state of the particle (relevant for e.g. partially stripped heavy ions).

N.B.: nested geometries are treated a a partition of space, so the deposited charge in a mother volume does not include contributions from its daughters. To score the daughter volume, a separate `DepositedChargeActor` should be attached to it.

N.B.: for repeated volume placements, the scored charge is the sum of the charges recorded across each repeated instance.

## Tests

A set of test scripts (`opengate/tests/src/actors/test099_*.py`) have been added. Testing this actor can be a bit challenging because of the interactions producing charged secondaries that can escape the scoring volume and bias the result. Tolerances have been added accordingly.

## Documentation

A section explaining the actor has been added to `docs/source/user_guide/user_guide_reference_actors.rst`.

Cheers :)

Marc


srmarcballestero:
Two new commits +  squashed follow-up fixes to address copilot review comments.

PR #1014, opened 2026-05-05
More robust uncertainty calculation

MartiDvrk:
This PR only affects the dose actor feature "stop simulation on uncertainty goal".

The average uncertainty of the image is calculated considering only the voxels that have an energy deposition larger than a threshold. Before, this threshold was defined as a fraction of the maximum energy deposition in the image. However, since this value can fluctuate a lot, especially in the early phase of the simulation, it is better to use the maximum of the N highest voxels, with N configurable and range checked.


MartiDvrk:
@yihan-jia

nkrah:
I made a minor patch and added updated the user guide. Good to go from my side. @MartiDvrk : should we merge this?

MartiDvrk:
It would be nice to merge, yes!

PR #1029, opened 2026-05-19
cleanup generic source initialization on python side

tontyoutoure:
1. remove redundant validation for direction
2. move confinement check into position validator

### DIFF
--- a/opengate/sources/generic.py
+++ b/opengate/sources/generic.py
@@ -126,6 +126,14 @@ class PositionValidator(UserInfoValidatorBase):
                 f"In {context_name}: 'rotation' must be convertible to a 3x3 matrix/array, "
                 f"but got: {b.rotation}"
             )
+
+        # check confine
+        if b.confine:
+            if b.type == "point":
+                warning(
+                    f"In {context_name}, "
+                    f"confine is used, while position.type is point ... really ?"
+                )
         return context_name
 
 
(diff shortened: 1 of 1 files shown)


dsarrut:
thanks @tontyoutoure ;) ! I agree that the validation is redundant (also done in the cpp side), but I think I prefer to keep the python side validation as the error message is easier to manage. Ok for moving the confinement check

tontyoutoure:
> thanks @tontyoutoure ;) ! I agree that the validation is redundant (also done in the cpp side), but I think I prefer to keep the python side validation as the error message is easier to manage. Ok for moving the confinement check

@dsarrut By redundant I mean the direction type check is done inside the DirectionValidator, at `generic.py:158`. This redundancy actually stop me from using generics' initialzation ,although I replement my own DirectionValidator, therefore I hope it get removed.

nkrah:
@tontyoutoure : Can you explain this "This redundancy actually stop me from using generics' initialzation ,although I replement my own DirectionValidator" please? Thanks

tontyoutoure:
@nkrah It's not a big deal actually. I'm writing a[ source ](https://github.com/tontyoutoure/opengate/blob/window_turbo_source/opengate/sources/windowturbosource.py)designed for variance reduction in pinhole SPECT system. The direction section of the user info is completely redefined:
```
def _wts_direction_parameters():
    return Box(
        {
            "a1": [],
            "a2": [],
            "b1": [],
            "b2": [],
            "plane_distance": [],
            "plane_phi": [],
            "init_sampling_count": 1000000,
            "init_number_of_threads": 0,
            "act_ratio": [],
            "max_solid_angle": [],
            "skip_mode": False,
        }
    )
```
therefore does not contains `type` in it and therefore cannot reuse `GenericSource`‘s initialization.

nkrah:
I see. And do I understand correctly that the new source inherits from GenericSource?

tontyoutoure:
@nkrah Yes, you're right. And it depends on the source visualization PR #1031 . Coding and testing is done and I'm grinding the documentation.

PR #1031, opened 2026-05-20
Source visualization

tontyoutoure:
Currently it's only working for generic source and it's derived classes, since the GateVSource does not provide methods to generate position sampling.

Docs and tests are also included.

image


tbaudier:
Hello @tontyoutoure 

Thank you for your contribution, it's very helpful. I modify a little bit to be able to call the visualization parameters in the way than the other parameters. I let you check the commit and resolve the conflict.

tontyoutoure:
@tbaudier I've tested merging into current master branch (commit edac4cd). It generates no conflict now.

tontyoutoure:
@tbaudier currently all sources will be defaultly visualized with `("red", 2000, 3)`. Do you think this is a good idea?

dsarrut:
Very good, thanks @tontyoutoure ! 
- Backend : Maybe the "RequestVisualization" and other visualisation related function, should be in the GateVSource base class. What do you think ? Hence, it should work with other type of sources. 
- Frontend : the visualisation options
    source.visualization.count = 2000
    source.visualization.color = "yellow"
    source.visualization.size = 2
is good from my side. Unsure about the default values, my first guess is that, indeed, we can enabled the source visualisation by default. 
- Tests : does it work with all visualization modes (qt, vrml, gdml ?). If not, we only need to document it.

tontyoutoure:
@dsarrut  I'm not sure moving `RequestVisualization` into `GateVSource` is a good idea. Simpling moving it can work, but the current function signature `void RequestVisualization(G4int count, py::object color, G4double size);` clearly indicate that it's a point cloud, which need position sampling, which `GateVSource` is not able to do. On the other hand since @tbaudier change the python api from a method of `GenericSource` to its attributes, I think renaming `RequestVisualization` into `InitializeVisualiztion` and call it in `InitializeUserInfo` is a better idea.

tontyoutoure:
@dsarrut I'm sure this don't work under vrml:
image

However I cannot make gdml visualization work on my machine, not even test004_simple_visu_gdml.py, only result in SegFault. Can you test it with test010_generic_source_visu.py?

tbaudier:
Hello @tontyoutoure 

I modified a little bit the code. The visualization of the sources should be done only if the visualization flag is set. If not it lead to an error before.

tontyoutoure:
@tbaudier Hi, I think the ffaa fix should be in other branch, correct? Not a big deal if they are merged though. Other things looks fine to me.

tontyoutoure:
@tbaudier @dsarrut I think it will be nice to have a unified color interface. The validate_color function is a nice place to begin with. I hope to add some function like `DictGetColourVec`/`DictGetColourVec` on the C++ side. There are two ways to do it imo:
1. convert all color (string/rgb/rgba) into rgba at python side, meaning `validate_color` need to return the converted value.
2. handle type conversion at C++ side, meaning a little bit more code.

What do you think?

tbaudier:
> @tbaudier @dsarrut I think it will be nice to have a unified color interface. The validate_color function is a nice place to begin with. I hope to add some function like `DictGetColourVec`/`DictGetColourVec` on the C++ side. There are two ways to do it imo:
> 
> 1. convert all color (string/rgb/rgba) into rgba at python side, meaning `validate_color` need to return the converted value.
> 2. handle type conversion at C++ side, meaning a little bit more code.
> 
> What do you think?

Good idea, maybe another PR can be done for that.
I merge this one

PR #1036, opened 2026-05-25
Window turbo source

tontyoutoure:
Window Turbo Source (a.k.a. VRT for multipinhole SPECT system) implementation with tests/documentations. Porting from Gate 9 version: https://github.com/tontyoutoure/Gate/tree/turbo_source

This should be merged after the source visualization PR #1031  is merged, since dependence exists.


dsarrut:
Very interesting! We should discuss more about this PR: the name "window turbo source" is probably not very well adapted ;), source visualisation is mixed here (should be separated), unclear how it compare to Angular Acceptance, etc. Lets keep in touch !

tontyoutoure:
@dsarrut I agree that stylewise, the interface should take some modifications and align with GenericSource.visualization.
Naming is always difficult. Do you have any recommendations?
I'd like to write a paper to discuss it, therefore comparing is to existing VRT is necessary, including AA.

PR #1038, opened 2026-05-25
porting forceWireframe and forceSolid

tontyoutoure:
Hi, I think these to options for visualizing volume in Gate 9 is quite handy when visualizing complicated structures, and take the liberty of porting them into opengate 10.

### DIFF
--- a/core/opengate_core/g4_bindings/pyG4VisAttributes.cpp
+++ b/core/opengate_core/g4_bindings/pyG4VisAttributes.cpp
@@ -17,5 +17,7 @@ void init_G4VisAttributes(py::module &m) {
       .def("SetColor",
            [](G4VisAttributes &va, G4double red, G4double green, G4double blue,
               G4double alpha) { va.SetColor(red, green, blue, alpha); })
-      .def("SetVisibility", &G4VisAttributes::SetVisibility);
+      .def("SetVisibility", &G4VisAttributes::SetVisibility)
+      .def("SetForceWireframe", &G4VisAttributes::SetForceWireframe)
+      .def("SetForceSolid", &G4VisAttributes::SetForceSolid);
 }
--- a/docs/source/user_guide/user_guide_how_to_convert_example_1.rst
+++ b/docs/source/user_guide/user_guide_how_to_convert_example_1.rst
@@ -101,6 +101,7 @@ GATE 9
     /gate/Spiral/setMaterial                Water
     /gate/Spiral/vis/setColor               yellow
     /gate/Spiral/vis/setVisible             1
+    /gate/Spiral/vis/forceWireframe
 
     # Insert at radius 0.00 mm and angle 0.00 degree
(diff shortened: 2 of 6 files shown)


tbaudier:
Hello @tontyoutoure 

I tried your PR, and I do not succeed to have a good results.
I started with the test misc/test004_simple_visu_qt.py and I added the line: (https://github.com/tontyoutoure/opengate/blob/refs/heads/source_visualization/opengate/tests/src/misc/test004_simple_visu_qt.py#L36)
```
waterbox.vis_type == "wireframe"
```
"wireframe" or "solid" I do not see any differences. Is it possible for you to propose a test like test004_simple_visu in which we can see the difference between wireframe and solid please?

tontyoutoure:
@tbaudier Hi, if you wrote = instead of == in your test and you just typoed here, I will try it myself couples of hours later.

tontyoutoure:
This is what I got running the test directly:



this is what I got after adding `waterbox.vis_type = "solid"` after line 36:

tbaudier:
Sorry, I did a mistake. I have the same results than you for qt. But I do not see any differences with vrml, and you? BTW can you add a test004_simple_visu_solid.py with solid option please?

tontyoutoure:
I'm not familiar with the vrml system. I will try looking into it but I can't guarantee the result.

I will add an test for the solid option. On the other hand, maybe call it `visu_type` can better align with other APIs? E.g. `sim.visu_type`

tontyoutoure:
@tbaudier I took a look at the vrml visu test and get the same result with you. I dig into Geant4 vrml visulization code, and found the following code at source/visualization/VRML/src/G4VRML2SceneHandlerFunc.icc:637
```c++
if(pVA)
{
  if(pVA->IsForceDrawingStyle() &&
     (pVA->GetForcedDrawingStyle() == G4VisAttributes::wireframe))
  {
    // wireframe style is assigned to this object
    flagWF = true;
  }
  else if(cur_view_param.GetDrawingStyle() == G4ViewParameters::wireframe)
  {
    // current viewing style is  wireframe
    flagWF = true;
  }
}
```
which means if you set the scene view style to wireframe, VRML visualizer will render all volume as wireframe (which means transparent). After changing view style, changing waterbox's style does gives correct result:

image

vs

image

However I don't find a way of running `/vis/viewer/set/style s` on Python side. The above image are generated with a hacky way. Maybe you can find a way get it working? I tried 
```
    sim.g4_commands_after_init=["/vis/viewer/set/style s", "/vis/viewer/flush"]
```
and it did not work.

I renamed the attribute from `vis_type` to `style` to align with Geant4 naming norms. A test is also added.

tbaudier:
Thank you for the renaming and the test. I also try on my side for vrml without success. Maybe we can add in the doc that it's only available for Qt.

tontyoutoure:
@tbaudier look good to me!

PR #1043, opened 2026-05-30
Fix the docs

djboersma:
My goal was to make the docs build (`make html`) with zero errors/warnings. Mission accomplished. However, I see that there is a lot of work on-going, currently, so 5 seconds after I file my PR new crashes and warnings will appear.

Some autodoc links/references (e.g. `PhaseSpaceActor.root_output`) were failing due to simple typos, others (about a dozen) because the corresponding `autoclass`, `automethod` or `autoproperty` call was missing. I added a mock docomentation of opengate_core.G4FieldManager in order to get rid of a sphinx build warning. Of course mock doc could be decorated with real documentation info about the methods, their args and return values, I might do that later, if it's useful.

I found an rst file describing the initialization of actors, which was not included in the TOC. I fixed its grammar and put it back. Please remove if there was a non-grammar reason for leaving it out.

I see that the `ProcessDefinedStepInVolumeAttribute` class (in `digitizers.py` is now obsolete and has been renamed to `ProcessDefinedStepInVolumeAttributeLegacy`. I see that there is some update about that in the docs, but trying to import that legacy class causes an exception to be thrown, so no `autoclass` should be included in the sphinx docs, to avoid that crash.

There is much more that can be improved on the docs, but let me just submit this, this is a useful chunck of work (I hope).

### DIFF
--- a/docs/source/developer_guide/developer_guide_code_structure.rst
+++ b/docs/source/developer_guide/developer_guide_code_structure.rst
@@ -308,3 +308,8 @@ inherits from another class, call the superclass:
 
 In particular, the GateObject superclass (and variants) implement a
 ``__str__()`` method which lists all user_info of the object.
+
+Reference
+---------
+
+.. autoclass:: opengate.base.GateObject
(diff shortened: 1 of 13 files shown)


tbaudier:
HEllo @djboersma 
Thank you very much, I merge

PR #1051, opened 2026-06-05
GammaAncestorAttribute 

dsarrut:
- new attribute to store the position or energy of the first gamma ancestor
@anetxe : no test yet, no doc. To be discussed


nkrah:
Nice. I was wondering: does this need to be specific to gammas, or can it be a more generic "ancestor" attribute where the user selects the particle?

PR #1057, opened 2026-06-12
windows-latest now is windows2025-vs2026

tbaudier:
no MSVC found during cibuild wheel
Try to generate the cache again to see if it resolve the problem

https://github.blog/changelog/2026-02-05-github-actions-early-february-2026-updates/#windows-server-2025-with-visual-studio-2026-image-now-available-for-github-hosted-runners

Same error than here: https://discourse.cmake.org/t/cmake-4-2-2-not-able-to-detect-vs2026-outside-developer-command-prompt/15478

### DIFF
--- a/.github/workflows/actions_build/action.yml
+++ b/.github/workflows/actions_build/action.yml
@@ -48,8 +48,8 @@ runs:
       uses: actions/cache@v4
       with:
         path: ~/software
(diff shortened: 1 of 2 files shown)


tbaudier:
Update cmake to have at least version 4.2.0:
https://discourse.cmake.org/t/could-not-find-any-instance-of-visual-studio/15377/3

Issue #1059, opened 2026-06-15
[bug] If total primary count in a run larger than 2^31, program will crash in the middle of simulation


tontyoutoure:
The title says it all. I think it has something to do with 
```
  oss << "/run/beamOn " << INT32_MAX;
```
in `GateSourceManager.cpp`. I tried to change `INT32_MAX` to a larger value, and noticed that the actual constrain come from `G4RunManager::BeamOn` uses `G4int` as its first parameter, so nothing can be done from the gate side. However a notice in documentation and/or crash before the simulation starts is still prefered IMO.

nkrah:
Thanks for pointing this out. This has caused issues on multiple occasions already. 
PR #1068 implements a little guarding mechanism to catch this annoying case and warn the user.

PR #1068, opened 2026-06-17
Warning int32 max

nkrah:
This PR introduces a check mechanism in the source management in GATE 10 to avoid that long simulations crash when the number of particles exceeds the internal Geant4 limit of INT32_MAX. This issue has been reported on multiple occasions, most recently in #1059.


nkrah:
One failing test at this point: 
actors/test097_pileup_mt.py
https://github.com/OpenGATE/opengate/actions/runs/27750878597/job/82109543942

Passes locally on my machine. I think this is unrelated. 

@gertvanhoey : what do you think?

PR #1072, opened 2026-06-20
make ProcessDefinedStepInVolumeAttribute able to follow multiple volumes

tontyoutoure:
For example, if a phantom is constituted with several parts, then it's prefered to count how much a particle is scattered in all those parts in total. Current `ProcessDefinedStepInVolumeAttribute` can only count how much the particle is scattered in those parts seperately, creating unnecessary c++ object footprint and enlarge the output file size. This PR is addressing this issue, making the `ProcessDefinedStepInVolumeAttribute` being able to follow a list of volumes and check whether a scatter happens in any of them.


tontyoutoure:
probably not a efficient way to implement.

PR #1076, opened 2026-06-25
Add surface_sphere source type for isotropic background simulations

cpommranz:
## Motivation

This PR adds support for generating particles on the surface of a sphere with directions sampled according to Lambert's cosine law and oriented toward the sphere interior, resulting in an isotropic radiation field within the sphere interior. This configuration is commonly used in astrophysical Geant4 applications to model an isotropic radiative background in space.

## Implementation

The implementation introduces a new source position type: `surface_sphere`.

I am aware that this is not a direct one-to-one mapping of the underlying Geant4 GPS configuration, where this source corresponds to the combination:

* /gps/pos/type: `Surface`
* /gps/pos/shape: `Sphere`

Ideally, these two concepts could be exposed independently. However, the current GATE source position types already combine the Geant4 `type` and `shape` parameters into a single user-facing option. For example:

* `sphere` corresponds to `Volume` + `Sphere`
* `disc` corresponds to `Beam` + `Circle`

Following this existing design pattern, introducing `surface_sphere` seemed to be the most consistent and least disruptive solution.

### DIFF
--- a/core/opengate_core/opengate_lib/GateGenericSource.cpp
+++ b/core/opengate_core/opengate_lib/GateGenericSource.cpp
@@ -395,12 +395,12 @@ void GateGenericSource::InitializeDirection(py::dict puser_info) {
   auto *ang = fSPS->GetAngDist();
   auto ang_type = DictGetStr(user_info, "type");
   fangType = ang_type;
-  std::vector<std::string> llt = {"iso", "histogram", "momentum", "focused",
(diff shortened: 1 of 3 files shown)


dsarrut:
very good, thanks a lot ! Maybe we should change the interface and follow G4 by separating type and shape ? Lets keep this version for the moment, but we should think about it

PR #1080, opened 2026-06-30
Implement order_coincidence_pairs function

ZainaHurani:
Added a function to reorder coincidence pairs based on specified criteria (time, distance, energy).


dsarrut:
thanks @ZainaHurani ! Can you please add a short test for this function to check that it works as expected (not urgent) ?

PR #1083, opened 2026-07-02
Enhance DepositedChargeActor with uncertainty statistics

srmarcballestero:
Enhance the DepositedChargeActor to include history-by-history uncertainty statistics for deposited charge. Add corresponding tests and documentation.


dsarrut:
thanks @srmarcballestero ;) ! It slightly unclear for me what happens when multi runs ? The uncertainty is only available for the last run ?

srmarcballestero:
> thanks @srmarcballestero ;) ! It slightly unclear for me what happens when multi runs ? The uncertainty is only available for the last run ?

Hi @dsarrut. As it is coded right now, the data available after a multi-run is a merged total, so the uncertainty refers to this post-simulation grand-total. I'll fix this to make per-run data available as well and reopen the PR :)

PR #1084, opened 2026-07-03
Reduce GlobalTime divergence of threads, reducing memory usage of act…

gertvanhoey:
`GateTimeSorter` is used by `GateCoincidenceSorterActor`, `GateDigitizerDeadTimeActor` and `GateDigitizerPileupActor` to sort digis according to `GlobalTime`. In multi-threaded simulations, if the `GlobalTime` diverges between threads, more digis need to be buffered before `GateTimeSorter` can guarantee that the digis are sorted. To reduce memory consumption, it is important to reduce `GlobalTime` divergence between threads. `GateTimeSorter` does this by having the fastest progressing thread do the actor's work, which slows this thread down relative to the other threads. 

The `GlobalTime` divergence can be reduced more by making the fastest thread `yield()` after it has executed the actor's work, giving more opportunity to the threads with slower `GlobalTime` increase. This call to `yield()` is what is being added in this PR.

### DIFF
--- a/core/opengate_core/opengate_lib/digitizer/GateTimeSorter.cpp
+++ b/core/opengate_core/opengate_lib/digitizer/GateTimeSorter.cpp
@@ -11,7 +11,7 @@
 #include "GateDigiCollectionManager.h"
 #include "GateHelpersDigitizer.h"
 #include <G4Threading.hh>
-#include <memory>
+#include <thread>
 #include <utility>
 
 GateTimeSorter::GateTimeSorter(const std::string &name) : fName(name) {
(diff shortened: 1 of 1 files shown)


dsarrut:
seems good ! Ready to merge ?

gertvanhoey:
I did some more testing and I am no longer convinced that this change always works. To be investigated.

gertvanhoey:
After closer investigation `thread::yield()` does not seem to help much to reduce `GlobalTime` divergence. A different PR will be made with a different approach, as discussed in https://github.com/OpenGATE/opengate/pull/1090.

gertvanhoey:
See previous comment for the reason why this PR is closed.

PR #1090, opened 2026-07-09
Timesorter improvement

gertvanhoey:
@dsarrut @nkrah @tbaudier 

GateTimeSorter has a mechanism for limiting the GlobalTime divergence between threads in multi-threaded simulations, with the goal of reducing the amount of memory needed for buffering digis during sorting. That mechanism gives more work to the thread that is most ahead in terms of GlobalTime, with the goal of slowing it down to give the other threads the opportunity to catch up.

It appears that there are two issues with this approach:

1. The mechanism is not always effective: depending on the setup of the simulation, the sorting work may only be a tiny fraction of the overall work done by Geant4 and GATE. So shifting that work to the fastest thread does not help a lot. Since the GlobalTime divergence is probabilistic, it may be that one execution of a simulation has a small thread divergence (making you think that the mechanism works), but when executing the same simulation again, the divergence might be a lot higher, consuming more memory.
2. When a simulation setup has more than one TimeSorter (e.g. a pile-up actor followed downstream by a coincidence actor), the behavior of the downstream TimeSorter becomes problematic because it will buffer many, if not all, digis until the end of the simulation, waiting to sort and process them until the very end of the simulation (or run). If the simulation is long, there may not be enough memory available to reach the end of the simulation. The reason for this problematic behavior is that the GlobalTime of digis observed by a thread is no longer an indication of the current GlobalTime in that thread, because digis can switch from one thread to another in the upstream TimeSorter.

What I propose to do:
* **This PR**: Remove the "fastest thread does the work" mechanism, which solves issue 2. If this could make it to the upcoming release, it would allow users to have multi-threaded simulations with combinations of dead time, pile-up and coincidence actors without being limited to rather short simulations.
* **Future PR**: A mechanism for reducing GlobalTime divergence is still needed, otherwise multi-threaded simulations would slowly keep growing in memory. As an alternative, I propose to impose a thread "barrier", which would mean that once in a while the fast threads wait for the slow ones to catch up. I have a working implementation of this, but it still needs tuning to make sure that the barrier is used frequently enough to keep memory usage low, but not too frequently to prevent that threads spend too much time waiting. This mechanism will probably not be finished and sufficiently tested before the upcoming release.

As illustration, this is the memory usage as a function of time for a multi-threaded simulation containing pile-up + dead time + coincidence actors, once when run from the current master branch, and once with the code from this PR (removal of the "fastest thread does the work" mechanism). As you can see, even after the change in this PR, memory is still slowly going up over time, which would be solved by the future PR.

image

### DIFF
--- a/core/opengate_core/opengate_lib/digitizer/GateTimeSorter.cpp
+++ b/core/opengate_core/opengate_lib/digitizer/GateTimeSorter.cpp
@@ -139,14 +139,9 @@ void GateTimeSorter::OnEndOfEventAction(std::function<void(void)> work) {
   // with a mutex, because all threads must copy their digis sequentially into
   // the same buffer.
   // 2. The actual time-sorting, followed by the work provided by the actor,
-  // happens next on two conditions: (i) a minimum number of ingestions
-  // has happened since the previous time sorting, and (ii) the current thread
-  // is the one with the most advanced GlobalTime value. If these conditions are
-  // not fulfilled, the execution of stage 2 is postponed until a later
-  // invocation of OnEndOfEventAction. These conditions avoid incurring the
-  // overhead of stage 2 after every ingestion of (typically very few) digis and
-  // make sure that the fastest progressing working threads do more work than
-  // the slower ones, which reduces GlobalTime divergence between threads.
+  // happens next when a certain number of ingestions has happened since the
+  // previous time sorting. This condition avoid incurring the overhead of stage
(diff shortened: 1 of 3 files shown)


dsarrut:
excellent @gertvanhoey ! Ready to be merged ?

gertvanhoey:
Yes, @dsarrut, ready to be merged.

PR #1097, opened 2026-07-19
Reduce memory consumption of multi-threaded simulations

gertvanhoey:
If a multi-threaded simulation contains an actor of which `output_filename` is not set (no root output), then the actor's output collection does not get cleared as frequently as intended, leading to increased memory consumption.

The condition for clearing the output collection is `event->GetEventID() % fClearEveryNEvents == 0`, but since events have globally unique IDs and are distributed to the working threads, every ID that is a multiple of `fClearEveryNEvents` is only seen by one thread.  In case of `N` threads, every thread clears its output collection on average once per `N * fClearEveryNEvents `, resulting in larger than intended momentary peak sizes of the digi collection. Moreover, a `clear()` of the underlying `std::vector`s sets their sizes to 0, but leaves their capacity unchanged. So the maximum size that a digi collection has reached at any point during the simulation determines how much memory it occupies going forward. The result is that memory consumption can continue to increase as long as the simulation is running.

The proposed change remembers per thread whether the digi collection clear for a given multiple of `fClearEveryNEvents ` has already happened.


gertvanhoey:
Rebased (x3)

PR #1105, opened 2026-07-27
Add GDML volume import support

qmisslin:
## Summary

- expose `G4GDMLParser` in `opengate_core`
- add a `GDMLVolume` type to import a complete GDML geometry subtree
- preserve the OpenGATE world and place the imported GDML root inside a selected mother volume
- support GDML setup selection, schema validation, name stripping, and parser overlap checks
- add user documentation and a regression test

## Motivation

OpenGATE currently supports exporting geometry to GDML for visualization, but does not provide a native volume type for importing an existing GDML geometry into a simulation.

This change allows externally generated Geant4 geometries to be reused directly inside an OpenGATE geometry hierarchy.

## Validation

- clean editable rebuild of `opengate_core`
- `G4GDMLParser` availability and construction verified
- GDML solids, materials, logical volumes, and placements imported successfully
- particle transport through the imported geometry verified
- `KillActor` attachment to the imported GDML subtree verified
- regression test `test113_gdml_volume.py` passes
- `pre-commit` hooks pass:
  - trailing whitespace
  - Black
  - clang-format

### DIFF
--- a/core/opengate_core/g4_bindings/pyG4GDMLParser.cpp
+++ b/core/opengate_core/g4_bindings/pyG4GDMLParser.cpp
@@ -1,0 +1,29 @@
+#include <pybind11/pybind11.h>
+
+#ifdef USE_GDML
+
+#include "G4GDMLParser.hh"
+
+namespace py = pybind11;
+
+void init_G4GDMLParser(py::module &m) {
+  py::class_<G4GDMLParser>(m, "G4GDMLParser")
+      .def(py::init<>())
+
+      .def(
+          "Read",
+          [](G4GDMLParser &parser, const G4String &filename, G4bool validate) {
+            parser.Read(filename, validate);
+          },
+          py::arg("filename"), py::arg("validate") = true)
+
+      .def("GetWorldVolume", &G4GDMLParser::GetWorldVolume,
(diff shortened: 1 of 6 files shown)


dsarrut:
You are too fast @qmisslin ;) ! The actions will probably fail because G4 is not installed with the gdml/XercesC but we will look at it.

qmisslin:
> You are too fast @qmisslin ;) ! The actions will probably fail because G4 is not installed with the gdml/XercesC but we will look at it.

Thanks David ! The CI logs seem to confirm your hypothesis.

`test113_gdml_volume.py` passes on the Ubuntu jobs, but it fails on all Windows jobs (Python 3.11 to 3.14) because `G4GDMLParser` is not exposed in the Windows `opengate-core` wheel.

The current Windows build script appears to compile Geant4 without `GEANT4_USE_GDML=ON` and without an explicit Xerces-C dependency, whereas the Linux and macOS builds already enable GDML.

The other Ubuntu failures come from `test111_jobs_run_sequential.py` and seem unrelated to this PR.

Would you prefer that we enable GDML/Xerces-C in the Windows build, or that the GDML regression test be skipped when the installed Geant4 does not provide GDML support?

PR #1112, opened 2026-08-06
EM Field Enhancement II: resolve field frames via the navigator, nested-volume fields, update defaults

srmarcballestero:
# Rationale

A GATE field is defined in the local frame of the volume it is attached to, so every `GetFieldValue` call has to convert a world point into that volume's frame and rotate the field vector back to world corrdinates before returning. Until now each field cached the world->local transform of every physical placement of its volume, and recovered "which placement am I in?" by testing the query point against each one in sequentally. Points outside all of them -which can happen, because of boundary overshoots by the integration machinery- fell back to the nearest placement, unless they exceeded a tolerance derived from an assumed maximum curvature radius of 100 m, in which case the run aborted.

That design had four problems:

- the tolerance rested on a hard-coded value;
- `G4VSolid::DistanceToIn(p)` returns a *safety*, i.e. an underestimate, so the
  "nearest placement" search can be not reliable;
- the cache must be refreshed by hand whenever the geometry moves between runs, and a
  missed refresh silently evaluates the field in the wrong frame;
- rebuilding the transforms assumed every ancestor is singly placed, which is wrong for a field volume nested inside a
  repeated ancestor.

# What changed

**Placement resolution.** `GateFieldBase` now stores the field's `G4LogicalVolume` and reads the transform per query from the navigation history: during tracking from the pre-step touchable, and outside tracking from the navigator, which is what field visualisation needs.

The lookup walks the history outwards from the deepest level to the first level placing
the field's volume. That matters because fields are attached with `forceToAllDaughters=True`, so a track inside a daughter is still steered by the mother's field and must use the mother's frame.

**Nested-volume fields,** `SetFieldManager(fm, forceToAllDaughters=True)`
overwrites a daughter's field manager unconditionally, so a field on a daughter was
silently discarded whenever it was added before its mother's. The documented override
behaviour therefore only worked for one of the two add orders. `VolumeEngine` now assigns fields
outermost-first, so the result no longer depends on the order `add_field` was called.

**Defaults.** The default values for the parameters controlling the numerical integration of the equation of motion and the estimation of boundary crossings have been updated to match the defaults in Geant4 v11.4.1.

# Tests and Documentation

Updated and added new tests to check the nested volume branch. 

Cheers,

Marc ;)

### DIFF
--- a/core/opengate_core/opengate_lib/GateElectroMagneticField.cpp
+++ b/core/opengate_core/opengate_lib/GateElectroMagneticField.cpp
@@ -8,13 +8,8 @@
 #include "GateElectroMagneticField.h"
 
 GateElectroMagneticField::GateElectroMagneticField(
-    G4ElectroMagneticField *inner, const G4VSolid *solid,
-    std::vector<G4ThreeVector> translations,
-    std::vector<G4RotationMatrix> rotations, double deltaChordMM)
-    : G4ElectroMagneticField(),
-      GateFieldBase(solid, std::move(translations), std::move(rotations),
-                    deltaChordMM),
-      m_inner(inner) {}
+    G4ElectroMagneticField *inner, const G4LogicalVolume *logicalVolume)
+    : G4ElectroMagneticField(), GateFieldBase(logicalVolume), m_inner(inner) {}
 
 void GateElectroMagneticField::GetFieldValue(const G4double Point[4],
                                              G4double *BEfield) const {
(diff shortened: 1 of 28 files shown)


dsarrut:
Thanks @srmarcballestero ;) ! Ready to be merged ?

srmarcballestero:
> Thanks @srmarcballestero ;) ! Ready to be merged ?

@dsarrut all good! :)

